PR #60, opened 2019-11-11
Monorepo

ysugimoto:
This PR makes this project as monorepo.

Planning: https://trello.com/c/GuGUFuDE/123-implement-monorepo-publishing-to-npm#comment-5dc57f8aa682847de4acd9ee

Note that this PR will changes the previous structure.

## What changes

- renamed packages with `tc-` prefix
- `test-helpers` also can be in the package (won't publish)
- resolved to import module path from relatives to package names except for old implementation (e.g. api/* and example-schema)
- passing tests
- configured publishing to npm module via CircleCI
- moved `tc-api-express/lib/request-context.js` to its module as `tc-api-express-logger` in order to avoid cyclic dependency
- moved neo4j related module in order to avoid cyclic dependency in `tc-test-helpers`, and would be better in `tc-api-db-manager` because they are relates to neo4j, I thought
  - `api-rest-handlers/lib/neo4j-model.js` -> `api-db-manager/lib/neo4j-model.js`
  - `api-rest-handlers/lib/neo4j-type-conversion.js` -> `api-db-manager/lib/neo4j-type-conversion.js`

## Need

- [ ] Works on Windows


api/server/routes/rest/lib/validation.js
```diff
@@ -2,7 +2,7 @@ const httpErrors = require('http-errors');
 const {
 	validators,
 	TreecreeperUserError,
-} = require('../../../../../packages/schema-sdk');
+} = require('@financial-times/tc-schema-sdk'); // eslint-disable-line import/no-extraneous-dependencies
```
ysugimoto: I'm not sure why eslint tells error but need disabling otherwise linting will fail

package.json
```diff
@@ -37,10 +37,15 @@
     "valid-url": "^1.0.9"
   },
   "devDependencies": {
+    "@financial-times/athloi": "github:ysugimoto/athloi",
```
ysugimoto: Temporary use my fork. need change to actual version after https://github.com/Financial-Times/athloi/pull/58 has been merged and released.

PR #67, opened 2019-11-18
Lists

wheresrhys:
# Why
It's been a long standing thing on the wishlist to support lists of strings (aka tags) and enums (aka multiple choice) stored in single properties on records. Combining the schema and api in one repo has revealed that this is actually really easy to do. Still the question of UI to solve (which will be harder for the tagging case (though fairly straightforward for multiple choice checkboxes). Not planning on merging this imminently as I don't believe in adding to the api without a plan for UI

# What
hasMany is permissable for enum and word fields in the schema. Once this is done, there's surprisingly little work needed in the apis:
- check for hasMany on all properties, not just rels, when generating graphql schema
- anticipate getting arrays of strings when validating hasMany properties
- diff of array before writing to these properties

### DIFF
--- a/example-schema/types/MainType.yaml
+++ b/example-schema/types/MainType.yaml
@@ -22,6 +22,11 @@ properties:
     type: Word
     description: Word description.
     label: Word label
+  someStringList:
+    type: Word
+    hasMany: true
+    description: String list description
+    label: String list label
   someBoolean:
     type: Boolean
     description: Boolean description.
@@ -30,6 +35,11 @@ properties:
     type: AnEnum
     description: Enum description.
     label: Enum label
+  someMultipleChoice:
+    type: AnEnum
+    hasMany: true
+    description: Multiple choice description
+    label: Multiple choice label
   someInteger:
     type: Int
     description: Int description.
--- a/packages/api-express/index.js
+++ b/packages/api-express/index.js
@@ -30,8 +30,7 @@ const getApp = async (options = {}) => {
 		restPath = '/rest',
 		restMiddlewares = [],
 	} = options;
-	updateConstraintsOnSchemaChange();
-	schema.init();
+
 
 	const router = new express.Router();
 	router.use(contextMiddleware);
@@ -47,7 +46,9 @@ const getApp = async (options = {}) => {
 		listenForSchemaChanges: updateGraphqlApiOnSchemaChange,
 	} = getGraphqlApi(options);
 
+	schema.init();
 	updateGraphqlApiOnSchemaChange();
+	updateConstraintsOnSchemaChange();
 
 	graphqlMethods.forEach(method =>
 		router[method](graphqlPath, graphqlMiddlewares, graphqlHandler),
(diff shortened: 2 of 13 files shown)


example-schema/types/MainType.yaml
```diff
@@ -22,6 +22,11 @@ properties:
     type: Word
     description: Word description.
     label: Word label
+  someStringList:
+    type: Word
+    hasMany: true
```
doramatadora: I am tempted to disagree with the use of `hasMany` here, because it implies **distinct** elements. We've been using it to describe the cardinality of relationships (this implies distinct elements), rather than to distinguish between primitive vs reference types. What about using a syntax like `Word[]` or `Array<Enum>`?

example-schema/types/MainType.yaml
```diff
@@ -30,6 +35,11 @@ properties:
     type: AnEnum
     description: Enum description.
     label: Enum label
+  someMultipleChoice:
+    type: AnEnum
+    hasMany: true
+    description: Multiple choice description
+    label: Multiple choice label
```
ysugimoto: Enum is suggestive `allowed fixed number` to me but the tests are actually using a string like 'First', 'Second'. Are these OK for us?  I can't find the actual value choice lists which this schema accepts on other place.

So I assume `AnEnum` field should have a field that declares user choosable values like `choise: ['First', 'Second']`

example-schema/types/MainType.yaml
wheresrhys: I did initially have 
```
type: 
  - Word
```

In the end I preferred hasMany because it introduces no new concepts. Having hand held a number of pull requests into the schema my instinct is that having multiple different ways of expressing the cardinality of a field will make things more confusing. 

From a maintenance point of view, it also keeps our code simpler if all we have to check is a boolean property vs checking for some sort of pattern/structure in the type. I suspect that even if we went for using a patten/structure for authoring the cardinality, we'd probably end up converting it into a boolean internally, and also having to store the primitive type/enum type in a new property too

i.e.

```
    def.hasMany = /[]$/.test(def.type)
    def.realType = def.type.replace(/[]$/, '')
```

example-schema/types/MainType.yaml
wheresrhys: The enums defined in the graphql api map directly to string representations of them stored in the database. I don't love how enums work in graphQL to be honest, but this is how it is, and it's consistent with other uses of enum. From wikipedia:

- a data type consisting of a set of named values called elements, members, enumeral, or enumerators of the type. The enumerator names are usually identifiers that behave as constants in the language

doramatadora:
What is the difference between an array of `Word` and an array of `AnEnum`? Why not just have the former, or `Enum`? It's not like we specify the acceptable enum values in the schema.

wheresrhys:
> It's not like we specify the acceptable enum values in the schema.

Ah but we do https://github.com/Financial-Times/treecreeper/blob/master/example-schema/enums.yaml

PR #69, opened 2019-11-21
Design custom relationship properties

ysugimoto:
~~# This PR is still work in progress~~

I defined relationships as root types and replace them.

## Changes

- Put new relationship types in `relationships/` directory
- Collect with `schema.types` and resolve as old structure (for backward-compatibility)

## ToDos

- [x] Transform to GraphQL schema with new structure and rich relationship property
- [x] Passing validator with new structure
- [x] Having backward compatibility
- [x] Adopt new relationship definition to current logics

~~Could you confirm my plan is going OK?~~

Here is actual graphql schema definition output
[defs.txt](https://github.com/Financial-Times/treecreeper/files/3896536/defs.txt)


packages/tc-schema-sdk/lib/updater.js
```diff
@@ -3,6 +3,14 @@ const fetch = require('node-fetch');
...
 const readYaml = require('./read-yaml');
 const { version: libVersion } = require('../../../package.json');
+const { compatBackward } = require('./compat-backward');
+
+// e.g HasChild -> HAS_CHILD
+const toSnakeUpperCase = name =>
+	name
+		.replace(/([A-Z])/g, '_$1')
+		.replace(/^_|_$/g, '')
+		.toUpperCase();
```
ysugimoto: This function will generate a relationship name automatically from it's rel name

example-schema/types/MainType.yaml
```diff
@@ -51,37 +51,23 @@ properties:
...
+    type: HasYoungerSibling
     description: Parents description.
     label: Parents label
-    direction: outgoing
-    hasMany: true
   olderSiblings:
-    relationship: HAS_YOUNGER_SIBLING
-    type: MainType
+    type: HasYoungerSibling
+    direction: to
```
ysugimoto: Declare the direction explicitly (because HasYoungerSibling indicates `MainType` both from and to)

example-schema/relationships/HasOlderSibling.yaml
```diff
@@ -0,0 +1,10 @@
+name: HasOlderSibling
```
wheresrhys: Would prefer this to be something like the following:

```
name: ReflexiveRichRelationship (https://www.merriam-webster.com/dictionary/reflexive)
relationship: HAS_RICHLY_RELATED_MAIN
properties:
  someProp:
    type: Word
    description: Word description.
    label: Word label
to:
  type: MainType
  hasMany: true
from:
  type: MainType
  hasMany: true
```

And also, for relationships to other types

```
name: RichRelationship 
relationship: HAS_RICHLY_RELATED_CHILD
properties:
  someProp:
    type: Word
    description: Word description.
    label: Word label
to:
  type: ChildType
  hasMany: true
from:
  type: MainType
  hasMany: true
```

example-schema/relationships/HasYoungerSibling.yaml
```diff
@@ -1,4 +1,5 @@
 name: HasYoungerSibling
+relationship: HAS_YOUNGER_SIBLING
```
wheresrhys: There should only be a sinlge object, which has both a from and a to. Maybe I need to explain in more detail

packages/tc-schema-sdk/lib/updater.js
```diff
@@ -41,6 +58,16 @@ class SchemaUpdater {
 			};
 		}
 		if (schemaData) {
+			if (backwardCompatibility) {
+				console.log(
+					`[SCHEMA] backward compatiliby enabled. SDK transforms to old model definition`,
+				);
+				schemaData.schema.types = compatBackward(
```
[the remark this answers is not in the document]
ysugimoto: Where is the best point to transform it? I think we should transform schema before set schema data as `this.rawData.set(schemaData)`.  Or it should be transformed on accessing data via `data-accessors`?

packages/tc-schema-sdk/sdk.js
```diff
@@ -21,6 +22,7 @@ class SDK {
 		this.getType = this.createEnrichedAccessor(type);
 		this.getTypes = this.createEnrichedAccessor(types);
 		this.getGraphqlDefs = graphqlDefs.accessor.bind(this);
+		this.getRelationships = this.createEnrichedAccessor(relationships);
```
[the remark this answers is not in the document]
ysugimoto: This method returns just relationship that has a type defined in the yaml.

This method is using at here: https://github.com/Financial-Times/treecreeper/pull/69/files#diff-c89555cf5b6547095947e2fdccdb578aR10 but as you mentioned, I should not divide types and relation

packages/tc-schema-validator/tests/type-test-suite.js
```diff
@@ -0,0 +1,369 @@
...
+		it('has to direction', () => {
+			expect(type.to).toMatchObject({
+				type: expect.any(String),
+				hasMany: expect.any(Boolean),
+			});
+		});
+		it('has a name', () => {
+			expect(typeof type.name).toBe('string');
+		});
+		it('may have a isMutal', () => {
```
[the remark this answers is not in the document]
ysugimoto: I think `isMutal` property can be omitted (default as false). therefore we should validate if it defined but we need infer from `from.type === to.type`

wheresrhys:
Here are the steps I would take to implement this (roughly in order). Perhaps you can point out which ones are problematic @ysugimoto 

1. Implement getRelationshipType accessor as documented here https://github.com/Financial-Times/treecreeper/pull/69#pullrequestreview-323682883
2. Implement getRelationshipTypes accessor 
3. Use getRelationshipTypes in graphql generation
4. Use getRelationshipType in validator
5. Add `{richRelationships: Bool (default false)}` option to getType & getTypes (we will need to define the output structure fro this... not sure we have done this yet)

ysugimoto:
@wheresrhys Thanks! it much helps me!

packages/tc-api-graphql/index.js
```diff
@@ -36,6 +36,7 @@ const getGraphqlApi = ({
 				{ event: 'GRAPHQL_SCHEMA_UPDATE_FAILED', error },
 				'Graphql schema update failed',
 			);
+			console.log(error);
```
wheresrhys:
```suggestion
```

packages/tc-schema-sdk/lib/updater.js
```diff
@@ -27,7 +44,10 @@ class SchemaUpdater {
 		if (schemaDirectory && !schemaData) {
 			schemaData = {
 				schema: {
-					types: readYaml.directory(schemaDirectory, 'types'),
+					types: [].concat(
+						readYaml.directory(schemaDirectory, 'types'),
+						readRelationships(schemaDirectory),
```
wheresrhys: can relationship types go in a separate top level property of `relationshipTypes`. They are rarely (never?) useful to have in a combined list, so woudl prefer to always keep them separate.

packages/tc-schema-sdk/lib/updater.js
```diff
@@ -2,6 +2,23 @@ const EventEmitter = require('events');
...
+		.replace(/^_|_$/g, '')
+		.toUpperCase();
+
+const readRelationships = schemaDirectory => {
+	if (!readYaml.isDirectory(schemaDirectory, 'relationships')) {
+		return [];
+	}
+	return readYaml.directory(schemaDirectory, 'relationships').map(rel => ({
+		...rel,
+		relationship: rel.relationship || toSnakeUpperCase(rel.name),
```
wheresrhys: Can you explain why this is needed? Is this supporting using the naming convention of the file to infer the relationship name, and therefore not expecting an explicit `relationship` property to exist? We don't support this for normal types, so I think we should either support it for both or for neither. 

I feel slightly in favour about always expecting the user to add a `relationship` property. It will be typed once, so not much effort to maintain, and makes things a bit less magic

packages/tc-schema-sdk/data-accessors/type.js
```diff
@@ -133,25 +133,53 @@ const getType = function(
...
+		Object.keys(typeSchema.properties).forEach(key => {
+			let definition;
+			try {
+				definition = this.getRelationshipType(typeSchema.name, key, {
+					primitiveTypes,
+					includeMetaFields,
+				});
+			} catch (error) {
+				definition = typeSchema.properties[key];
+			}
```
ysugimoto: In these lines, we can accept both of rich relationship and current implementation

packages/tc-schema-sdk/data-accessors/relationship-type.js
```diff
@@ -0,0 +1,141 @@
...
+	rootType,
+	property,
+	{ name, from, to, relationship, properties },
+) => {
+	const { direction } = property;
+	let relationshipTo;
+	let hasMany;
+	let dest;
+	if (from.type === to.type && !direction) {
+		throw new TreecreeperUserError(
```
wheresrhys: shouldn't this check be in the tc-schema-validator library? In general, the sdk does not validate that the schema rawData is of the expected structure. If you need to throw here in order to prevent the code later on executing successfully (but making bad decisions) throw a plain error, not a TreecreeperUserError (which are reserved for when a user passes bad data to the api, not when the maintainer of the application has misconfigured something)

packages/tc-schema-sdk/data-accessors/relationship-type.js
```diff
@@ -0,0 +1,141 @@
...
+) {
+	const relationshipType = getFromRawData(
+		rootType,
+		propertyName,
+		this.rawData,
+	);
+	if (!('properties' in relationshipType)) {
+		relationshipType.properties = {};
+	}
+	if (includeMetaFields && Object.keys(relationshipType.properties).length) {
```
wheresrhys: can some of these bits be extracted into shared library files used by relationship-type and type?

packages/tc-schema-sdk/data-accessors/relationship-type.js
```diff
@@ -0,0 +1,141 @@
...
+			const cloned = { ...def };
+			if (primitiveTypes === 'graphql') {
+				cloned.type = primitiveTypesMap[def.type] || def.type;
+			}
+			if (def.pattern) {
+				cloned.validator = this.getStringValidator(def.pattern);
+			}
+			return [name, cloned];
+		})
+		.filter(entry => !!entry)
```
wheresrhys: is this line needed?

packages/tc-schema-sdk/data-accessors/relationship-types.js
```diff
@@ -0,0 +1,40 @@
+const type = require('./type');
+
+module.exports = {
+	// todo move to object rest parameters for options when upgrading node
+	accessor(options = {}) {
+		const types = this.rawData.getTypes();
+
+		// Find root type names of rich relationships (newer designed)
+		const rootRichRelationshipTypeNames = types
```
wheresrhys: See comment elsewhere about keeping relationship types separate from types. Then this duck-typing won't be necessary

packages/tc-schema-sdk/data-accessors/relationship-types.js
```diff
@@ -0,0 +1,40 @@
+const type = require('./type');
+
+module.exports = {
+	// todo move to object rest parameters for options when upgrading node
+	accessor(options = {}) {
+		const types = this.rawData.getTypes();
```
wheresrhys: Question - does this return only relationshipTypes that have been user define in the schema, or does it also return ones which are generated from analysing relationship properties on types. My instinct is to return autogenerated ones too.

packages/tc-schema-sdk/data-accessors/type.js
```diff
@@ -133,25 +133,53 @@ const getType = function(
...
-				if (def.hidden) {
-					delete typeSchema.properties[propName];
+		Object.keys(typeSchema.properties).forEach(key => {
+			let definition;
+			try {
+				definition = this.getRelationshipType(typeSchema.name, key, {
+					primitiveTypes,
+					includeMetaFields,
+				});
+			} catch (error) {
```
wheresrhys: What are the scenarios when this errors?

packages/tc-schema-sdk/data-accessors/type.js
```diff
@@ -133,25 +133,53 @@ const getType = function(
...
 			}
 		});
 	} else {
-		Object.entries(typeSchema.properties).forEach(([propName, def]) => {
-			if (def.relationship) {
-				delete typeSchema.properties[propName];
+		Object.keys(typeSchema.properties).forEach(key => {
+			let definition;
+			try {
+				definition = this.getRelationshipType(typeSchema.name, key, {
```
wheresrhys: here you are using it to check if a property is a relationship. Is it worth implementing `isRelationship` which (once you've split realtionshipTypes into a separate array to types in the rawData) can just check to see if it can find the relationship type in the array?

packages/tc-schema-sdk/data-accessors/graphql-defs.js
```diff
@@ -288,15 +296,19 @@ module.exports = {
...
 		scalar Time
 	`;
-		const typeDefinitions = types.map(printTypeDefinition);
+		const typeDefinitions = types
+			.filter(({ from, to }) => !from && !to)
+			.map(printTypeDefinition);
 
 		const relationshipTypeDefinitions = printRelationshipTypeDefinitions(
-			types,
+			this.getRelationshipTypes(retrieveOptions),
```
wheresrhys: does this still create relationship types even when their is not a user defined yaml file?

packages/tc-schema-sdk/lib/updater.js
ysugimoto: Refer to here https://trello.com/c/xZ6SbEGP/154-design-representations-of-custom-relationship-properties#comment-5dcd8ad9df28d53b62813a4f - `relationship` property in the root relationship type may not define, so I intended to attach it from type name e.g `HasChild` -> `HAS_CHILD`.

But I agree with we always expect the user adds  `relationship` property, and it should validate on tc-schema-validator

packages/tc-schema-sdk/lib/updater.js
ysugimoto: Then can we define root relationship types in top level like:

```
schema: {
  types: ...,
  relationships: ...,
  ...,
}
```

I thought that the schema structure shouldn't change for backward compatibility.
But if I can do that, data accessors can be more simpler.

packages/tc-schema-sdk/data-accessors/type.js
ysugimoto: Yes - current this data-acceccor is working based on combined types array, so when we move relationship types toward other root property, this logic can be simpler (maybe we can check without calling `this.getRelationshipType`)

packages/tc-schema-sdk/data-accessors/type.js
ysugimoto: I implemented `getRelationshipType` accessor as same as `getType()`, which throws error if type is not found.
So we should catch that error and it reckons `relationship is not found` - but we can return `null` and compare -- it is more clearer

packages/tc-schema-sdk/data-accessors/relationship-types.js
ysugimoto: No in here, but actually returns relationships which resolved from type properties on following logic -- here: https://github.com/Financial-Times/treecreeper/pull/69/files/05f4aff2c1cece2790f97b9d69aa4a7fe31a48df#diff-dc4679c40157e5b6ce606145a4df7c11R13-R34

packages/tc-schema-sdk/data-accessors/relationship-type.js
ysugimoto: No.. I'll remove it

packages/tc-schema-sdk/data-accessors/relationship-type.js
ysugimoto: That's a good idea. I'll do that!

packages/tc-schema-sdk/data-accessors/graphql-defs.js
ysugimoto: Yes, `getRelationshipTypes` returns all resolved relationships in type properties with a bit more information e.g from, to, etc. Therefore we trust that `getRelationshipTypes()` returns _all_ relationships in user-defined type properties

packages/tc-schema-sdk/lib/updater.js
```diff
@@ -28,6 +35,7 @@ class SchemaUpdater {
 			schemaData = {
 				schema: {
 					types: readYaml.directory(schemaDirectory, 'types'),
+					relationshipTypes: readRelationshipTypes(schemaDirectory),
```
wheresrhys: for consistency, could `readYaml.directory` be implemented so it returns `[]` if the directory is missing, then here we could have `relationshipTypes: readYaml.directory(schemaDirectory, 'relationships')`

packages/tc-schema-validator/tests/types.js
```diff
@@ -1,386 +1,53 @@
...
-										/^incoming|outgoing$/,
-									);
-								});
-								it('may be hidden', () => {
-									if (config.hidden) {
-										expect(config.hidden).toBe(true);
-									}
-								});
+describe('validate types', () => {
+	types.filter(({ from, to }) => !from && !to).forEach(typeTestSuite);
```
wheresrhys: I don't understand why these filters are still here now that types and relationshipTypes are split into different getters?

packages/tc-schema-sdk/data-accessors/relationship-types.js
wheresrhys: ok I think I understand.

packages/tc-schema-sdk/data-accessors/relationship-types.js
```diff
@@ -0,0 +1,32 @@
...
+		const relationshipTypes = this.rawData.getRelationshipTypes();
```
wheresrhys:
```suggestion
		const userDefinedRelationshipTypes = this.rawData.getRelationshipTypes();
```
This makes it a bit easier to follow I think

packages/tc-schema-sdk/data-accessors/relationship-types.js
```diff
@@ -0,0 +1,32 @@
+const type = require('./type');
+
+module.exports = {
+	// todo move to object rest parameters for options when upgrading node
```
wheresrhys: can actually do this now - all apps using the sdk run in node 10 or higher

packages/tc-schema-sdk/data-accessors/relationship-types.js
```diff
@@ -0,0 +1,32 @@
...
+	accessor(options = {}) {
```
wheresrhys: I would add a comment at the top of this function explaining that it gets _all_ relationshipTypes - a mixture of user defined and automatically generated

packages/tc-schema-sdk/data-accessors/relationship-types.js
```diff
@@ -0,0 +1,32 @@
...
+		const types = this.rawData.getTypes();
+		const relationshipTypes = this.rawData.getRelationshipTypes();
+
+		// Find relationships from type properties
+		return types.reduce((relationships, { name, properties }) => {
+			Object.entries(properties).forEach(
+				([propName, { relationship, type: typeName }]) => {
+					if (
+						relationship ||
```
wheresrhys: I do wonder if we should expect type properties that reference relationship types to specify `isRelationship: true`. It would simplify the logic here (and maybe elsewhere?), and might make the schema easier to understand (or maybe not - I'm really not sure)

On the other hand, it's more future proof if we expect less in the yaml files now, and can add it in later (does not need to be a breaking change if we're careful about rolling out library updates before actually changing the schema)

What do you think?

packages/tc-schema-sdk/data-accessors/relationship-type.js
```diff
@@ -0,0 +1,129 @@
...
+		name,
+		relationship,
+		hasMany,
+	};
+};
+
+const getFromTo = (direction, rootType, otherType) =>
+	direction === 'outgoing' ? [rootType, otherType] : [otherType, rootType];
+
+const getFromRawData = (rootType, propertyName, rawData) => {
```
wheresrhys: I'm just about able to follow what's going on here. I think this file could benefit from being split into more, smaller functions and variables with longer names to make it clear what the logical steps are e.g.

```
let richRelationshipDefinition = getUserDefiniedRichRelationship(...)

if (!richRelationshipDefinition) {
   richRelationshipDefinition = constructRichRelationshipDefinition(...)
}
```

packages/tc-schema-validator/tests/types.js
ysugimoto: You are right. I should access each via `sdk.rawData.getTypes()` and `sdk.rawData.getRelationshipTypes()` without filtering

packages/tc-schema-sdk/data-accessors/relationship-types.js
ysugimoto: I agree, and if we assign `isRelationship: true` at each type properties, it should be done at `getType()`, I think.

However, now we're using `isRelationship` field in here - https://github.com/Financial-Times/treecreeper/blob/master/packages/tc-schema-sdk/data-accessors/type.js#L143 so i'll try this logic being simply with carefully. Probably above logic indicates that the property type is relationship explicitly so we can use it.

packages/tc-schema-sdk/lib/updater.js
```diff
@@ -27,7 +27,14 @@ class SchemaUpdater {
 		if (schemaDirectory && !schemaData) {
 			schemaData = {
 				schema: {
-					types: readYaml.directory(schemaDirectory, 'types'),
+					types: [].concat(
+						readYaml.directory(schemaDirectory, 'types'),
+						readYaml.directory(schemaDirectory, 'relationships'),
```
wheresrhys: Only 1 question from me now. Why is it needed to include relationships in here. Would prefer to keep types 100% separate from relationships in these rawData lists. What are you using this combined list for?

ysugimoto: Ah, I followed your comment: https://github.com/Financial-Times/treecreeper/pull/69#discussion_r352167849 but maybe I misunderstand. this is not necessary, so I'll remove this line

PR #70, opened 2019-11-22
Schema versioning

wheresrhys:
# Why
Decoupling schema content from library version means we need a new mechanism to identify when schema has changed

# What
This changes to use
- uses library version for url
- generates a hash of schema contents for schema version


wheresrhys:
I'm now thinking about removing the treecreeper version from the url. If we ever need to introduce breaking changes to the schema design, that can be reintroduced then, but in the immediate future what's more likely is to make breaking changes to how the packages interact, so always being forced to update all the applications just so they can point at the latest schema feels like it'd be an annoyance. We will always have this code to reintroduce as and when we need to version urls, but for now it feels unnecessary

PR #73, opened 2019-11-25
Rebuild2

wheresrhys:
# Why
Need to see if we can actually replace the existing api with treecreeper parts

# What
Big delete of old API code. Uncovered a handful of bugs in the packages, all of which are now fixed

## TODO
- [x] Test in tc-api-express that HEAD request error returns a Debug-Error header


wheresrhys:
I think this is too big to review, so gonna go ahead and merge

PR #77, opened 2019-11-26
Implementing diffing relationship properties before writing

fenglish:
This PR includes

- [x] **Normalise relationship value**
Relationship value could be
 => String(code)
 => Object
 => an Array of Strings(codes) / Objects(code and properties) / Mixed
Add a function to  normalise them into an Array of Objects. 
It will make easier when checks diffs.

- [x] **Diff detection for `String` relationship property**

[The trello ticket](https://trello.com/c/g2Czl4uZ/175-implementing-diffing-relationship-properties-before-writing)

---------

**These implementation hasn't done by this PR**
(I will create another PR to implement them)
- diff detection for `Datetime` relationship property
- `_updatedBy...` should update whenever a relationship property is changed


packages/tc-api-rest-handlers/lib/relationships/input.js
```diff
@@ -125,9 +195,34 @@ const containsRelationshipData = (type, payload = {}) => {
 		.some(([propName]) => propName in payload || `!${propName}` in payload);
 };
 
+// relationship value could be just a String(code), an Object or an Array of strings(codes) / objects(code and properties) / mixed
+// this function is to normalise them into an Array of objects
+const normaliseRelationshipProps = (type, body) => {
+	const relationships = getWriteAndDeleteRelationships({ type, body });
```
wheresrhys: having read the code and seen what this is doing, I think... hmmm... it is definitely a case of "Naming things is hard!"... but I will have a think about what might be clearer names.

packages/tc-api-rest-handlers/lib/relationships/input.js
```diff
@@ -33,24 +72,42 @@ const isDeleteRelationship = type => {
 		(isRelationship(propName) && codes === null);
 };
 
-const findActualDeletions = initialContent => ([propName, codesToDelete]) => {
+const isWriteAndDeleteRelationship = type => {
```
wheresrhys:
```suggestion
const containsRelationshipDefinitions = type => {
```

packages/tc-api-rest-handlers/lib/relationships/input.js
```diff
@@ -106,8 +163,21 @@ const getRelationships = ({ type, body = {} }, reduce = true) => {
 		: newRelationships;
 };
 
+const getWriteAndDeleteRelationships = ({ type, body = {} }, reduce = true) => {
```
wheresrhys:
```suggestion
const getExplicitlySpecifiedRelationships = ({ type, body = {} }, reduce = true) => {
```

packages/tc-api-rest-handlers/lib/relationships/input.js
wheresrhys: Have left some suggestions. 

But another - maybe simpler - way of dealing with this could be to not bother with filtering out relationshipProperties which just use `null`. So you can iterate through ALL relationship properties that are in the body, but just add a rule in this function that if the value is `null`, then return it without modification?

PR #78, opened 2019-11-26
use official athloi

ysugimoto:
`@finacial-times/athloi` has been released with including my changes, so now we can use it.

Trello:
https://trello.com/c/YU6nubN5/181-use-official-athloi-for-managing-monorepo

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ verify:
 install: monorepo install-treecreeper
 
 monorepo:
-	npm install ysugimoto/athloi
+	npm install @financial-times/athloi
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -34,8 +34,8 @@ verify:
 install: monorepo install-treecreeper
 
 monorepo:
-	npm install ysugimoto/athloi
-	npx athloi exec --concurrency 10 -- npm install --no-package-lock
+	npm install @financial-times/athloi
+	npx athloi exec --concurrency 1 -- npm install --no-package-lock
```
ysugimoto: We should set the concurrency as 1 otherwise raise an error of resolving each package dependencies

PR #81, opened 2019-11-27
Salesforce sync

wheresrhys:
# Why
Biz Ops updates salesforce whenever a new system is created. So we need a mechanism whereby an application using treecreeper is able to listen for data changes in order to carry out supplimentary actions.

# What
I started off trying to hook into the same mechanism as kinesis. This led to an epiphany that really what we need is to turn the app into an event emitter. So changes in this PR are:
- changing the logging of changes to kinesis to just be an event emitter used by tc-api-rest-handlers, and exposed on tc-api-express
- Removed tc-api-publish
- Moved kinesis adaptor into the /api directory
- Copied the salesforce sync code and tests from biz-ops-api into here


packages/tc-api-rest-handlers/lib/events/index.js
```diff
@@ -66,11 +87,16 @@ const logChanges = (
 		throw new Error(message);
 	}
 
-	const events = makeEvents(action, entity, relationships);
-	const publisher = createPublisher(adaptor);
-	publisher.publish(...events);
+	makeEvents(action, entity, relationships).forEach(event =>
+		emitter.emit(event.action, {
```
ysugimoto: Do we need one more property like `event.kind` in order to be easy to distinguish for subscribers which event should handle or not? Eventually, some other event-driven module will use this.

wheresrhys: what value would `event.kind` contain? `action` is already a verb, which suits being a name for an event

wheresrhys:
big build problems in CI - closing this PR while I resolve

PR #85, opened 2019-11-27
Run absorb with transaction

ysugimoto:
## Why

In `absorb` handler, we should execute queries in a transaction in order to rollback the absorbed node and removed relationships

## What

I add the `executeQueryWithTransaction` function and export it via `tc-api-db-manager`, and `tc-api-rest-handlers` use it.

In order to run multiple queries at once in same transaction scope, I declared `executeQueryWithTransaction()`  signature which accepts multiple query sets and return results as an array. see below of usage -- it is wrapped via `neo4j-model.js` to add biz ops enhancements.

```
// in tc-api-rest-handlers
const { executeQueryWithTransaction } = require('./libs/neo4j-model');

const transQueries = [
  { query: 'SOME DELETE QUERY', parameters: {...} },
  { query: 'SOME MERGE QUERY', parameters: {...} },
];

const [deleteResult, mergeResult] = executeQueryWithTransaction(...transQueries);
console.log(mergeResult.toJson({ type: 'MainType', ... });
```

And I adopt this to absorb handler, tests still pass.

### DIFF
--- a/packages/tc-api-db-manager/db-connection.js
+++ b/packages/tc-api-db-manager/db-connection.js
@@ -90,7 +90,7 @@ module.exports = {
 
 		return executeQuery;
 	},
-	executeQueryWithTransaction: async (...queries) => {
+	executeQueriesWithTransaction: async (...queries) => {
 		const session = originalSession.apply(driver, []);
 		const result = await session.writeTransaction(async tx =>
 			Promise.all(
--- a/packages/tc-api-db-manager/db-connection.js
+++ b/packages/tc-api-db-manager/db-connection.js
@@ -15,48 +15,61 @@ const driver = neo4j.driver(
 
 const originalSession = driver.session;
 
+const runQueryWithMetrics = async ({
+	session,
+	runner,
+	isTransaction = false,
+}) => {
+	const start = Date.now();
+	metrics.count('neo4j.query.count');
+	let isSuccessful = false;
+	try {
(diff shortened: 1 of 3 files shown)


packages/tc-api-db-manager/db-connection.js
```diff
@@ -28,37 +28,49 @@ const getTimeoutRacePromise = timeout =>
...
+		const result = TIMEOUT
+			? await Promise.race([runner(), getTimeoutRacePromise(TIMEOUT)])
+			: runner();
+
+		isSuccessful = true;
+		return result;
+	} finally {
+		if (!isTransaction) {
+			session.close();
+		}
```
ysugimoto: In the transaction, the session must not close here. the session must be closed after all queries have been run because neo4j-driver will do commit/rollback with using this session

packages/tc-api-rest-handlers/absorb.js
```diff
@@ -192,13 +192,33 @@ const absorbHandler = ({ documentStore } = {}) => async input => {
...
+				nodeType,
+				absorbedCode,
+				code,
+				metadata,
+			}),
+		);
+
+		// Entire queries which affects current node should run in transaction.
+		// So we stack queries and execute all at once
+		const results = await executeQueryWithTransaction(...queries);
```
wheresrhys: executeQueriesWithTransaction

PR #93, opened 2019-12-03
Adopt runbook.md parser into treecreeper

ysugimoto:
This PR imports runbook.md parser into treecreeper package as `tc-markdown-parser`.

## TODOs

- [x] Copy libraries into this project and define package as `@financial-times/tc-markdown-parser`
- [x] define parser factory function like `getParser(scheme, options)`
- [x] adopt tests (a few tests which relate to schema definition)
- [x] fill tests against example-schema

## Parser factory function signature

```
getParser({
   type, // the type to parse into
   h1Property = 'name' // the property to use the h1 property for
   1stParagraphProperty = 'description' // the property to use the first paragraph for
   fieldWhitelist
   fieldBlacklist
})
```

and usage:

```
const getParser = require('@financial-times/tc-markdown-parser');
const schema = require('@financial-times/tc-schema-sdk');

schema.init({ ... });

(async () => {
  // Don't need to wait schema is ready because this package will call schema.ready() internally
  const parser = getParser(schema, { type: 'MainType' });
  const parsed = parser.parseRunbookString(...);
  console.log(parsed);
})();
```

Currently, I mark some tests `skip` for passing tests with example-schema (see comments below)


packages/tc-markdown-parser/__tests__/property-value.spec.js
```diff
@@ -0,0 +1,302 @@
...
+		### hello
+	`);
+
+	expect(data).toEqual({
+		name: 'name',
+		someDocument: '# hello',
+	});
+});
+
+test.skip('last review date is banned', async () => {
```
ysugimoto: because `lastReviewDate` is not defined in MainType

packages/tc-markdown-parser/__tests__/property-value.spec.js
```diff
@@ -0,0 +1,302 @@
...
+		* ft-app-fruitcake
+		* apple-quicktime
+	`);
+
+	expect(errors).toHaveLength(0);
+	expect(data.children).toEqual(['chee-rabbits']);
+	expect(data.youngerSiblings).toEqual(['ft-app-fruitcake', 'apple-quicktime']);
+});
+
+test.skip('dependents is banned', async () => {
```
ysugimoto: because `dependents` is not defined in MainType

packages/tc-markdown-parser/example.js
```diff
@@ -0,0 +1,37 @@
...
+	// const { data, errors } = await parser.parseRunbookString(mdString);
+	// console.log(data, errors);
+	const processor = await unified()
+		.use(remarkParse)
+		.use(someOriginalTransformer)();
+
+	const parsed = await processor.parse(mdString);
+	await processor.runSync(parsed);
+	//console.log(JSON.stringify(tree, null, '  '));
+})();
```
ysugimoto: This file is just copied original. Need fix or remove

packages/tc-markdown-parser/lib/tree-mutators/validate-bizops-properties.js
```diff
@@ -0,0 +1,49 @@
+const visit = require('unist-util-visit-parents');
+const convertNodeToProblem = require('./convert-node-to-problem');
+
+const excludedProperties = [
```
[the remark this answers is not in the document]
ysugimoto: I see, and when blocklist is supplied as config, what should we deal with that config?

1. concat with default setting as defined in here
2. should be overridden with supplied config

packages/tc-markdown-parser/__tests__/property-value.spec.js
```diff
@@ -0,0 +1,302 @@
...
+
+	expect(errors).toHaveLength(1);
+
+	const [{ message }] = errors;
+
+	expect(message).toContain("mario's birthday");
+	expect(message).toContain('Date');
+});
+
+test('empty sections are an error', async () => {
```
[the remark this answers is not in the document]
ysugimoto: Does it means:

```
# some string

<!-- this is comment -->
```

should not be an error and should ignore in section value? I'll be able to do that.

packages/tc-markdown-parser/__tests__/property-value.spec.js
wheresrhys: can delete this test

packages/tc-markdown-parser/__tests__/property-value.spec.js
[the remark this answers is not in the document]
wheresrhys: If you can see how to, yes that would be a great improvement

packages/tc-markdown-parser/lib/tree-mutators/validate-bizops-properties.js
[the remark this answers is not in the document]
wheresrhys: nope - no default config. The default config can continue to live in the runbook.md app, and will be passed in to this new shared library

packages/tc-markdown-parser/__tests__/property-name.spec.js
```diff
@@ -0,0 +1,37 @@
...
+		# name
+
+		## Contains big monkey
+		yes
+	`);
+
+	expect(errors).toHaveProperty('length', 1);
+	expect(data).toEqual({
+		name: 'name',
+	});
```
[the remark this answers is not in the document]
ysugimoto: Now parser will overwrite property value from the after one, so I'll create validator to validate same property name is defined in markdown

packages/tc-markdown-parser/lib/markdown-parser.js
```diff
@@ -0,0 +1,97 @@
...
+const stringifyBoast = require('./unist-stringifiers/stringify-boast');
+
+/* @param schema: Treecreeper schema singleton */
+const unifiedProcessor = function(
+	schema,
+	{
+		type,
+		h1Property = 'name',
+		firstParagraphProperty = 'description',
+		// fieldWhitelist = [], FIXME: usage?
```
wheresrhys: we can probably leave it out actually. We have no use case for whitelist yet

packages/tc-markdown-parser/__tests__/property-value.spec.js
```diff
@@ -0,0 +1,330 @@
...
+		},
+		{
+			line: 5,
+			message: 'property "anotherString" has no value',
+		},
+	]);
+});
+
+describe('dealing with html comments', () => {
+	it('html comments only in section should not be an error but be omitted in result', async () => {
```
wheresrhys: Is there a test somewhere that checks that the following is invalid?

```
			# name
			## some string
			
			## some document
			
```

packages/tc-markdown-parser/__tests__/property-name.spec.js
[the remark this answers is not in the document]
ysugimoto: Oops, this has not implemented yet 😢 I'll do

packages/tc-markdown-parser/lib/tree-mutators/apply-blacklist.js
```diff
@@ -0,0 +1,26 @@
...
+	'sosTrafficLight',
+	'stakeholders',
+	'updatesData',
+	'dataOwner',
+	'gdprRetentionProcess',
+	'gdprErasureProcess',
+];
+
+const applyBlacklistProperties = (additionalBlacklistProperties = []) =>
+	defaultBlacklistProperties.concat(additionalBlacklistProperties);
```
wheresrhys: This doesn't need default blacklist. Should default to [] if none provided by user (runbook.md should take care of defining the blacklist we've been using so far)

packages/tc-markdown-parser/lib/tree-mutators/create-treecreeper-name-node.js
```diff
@@ -0,0 +1,61 @@
+const { selectAll } = require('unist-util-select');
```
wheresrhys: maybe this would be better named as create-treecreeper-title-node.js, as this takes the title of the doc and writes it into a field (which maybe should default to code, actually, and runbook.md can configure it to write to name instead)

packages/tc-markdown-parser/__tests__/property-value.spec.js
```diff
@@ -0,0 +1,302 @@
...
+});
+// There are currently no date fields permitted in runbook.md; lastServiceReviewDate is reserved for Ops manual entry
+test('date fields are coerced to iso strings', async () => {
+	const naiveJavaScriptIsoStringRegex = /^\d{4}(?:-\d{2}){2}T(?:\d{2}:){2}\d{2}\.\d{3}Z$/;
+	const { data } = await parser.parseRunbookString(here`
+		# name
+
+		## some date
+
+		July 21 2018
```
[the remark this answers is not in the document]
ysugimoto: Currently, the `Date` type accepts `Acceptable string as new Date(string)`, https://github.com/Financial-Times/treecreeper/blob/import-runbook-md-parser-into-treecreeper/packages/tc-markdown-parser/lib/property-coercers/index.js#L93-L94 so we follow the node runtime spec...

packages/tc-markdown-parser/lib/tree-mutators/apply-blacklist.js
ysugimoto: Ah, sorry I forgot to remove this

packages/tc-markdown-parser/__tests__/property-value.spec.js
ysugimoto: Just above tests https://github.com/Financial-Times/treecreeper/pull/93/files/acb8d01b66915104ad41817ec0dad960cfd0b383#diff-5624ce41fc406050be2a3489982935acR267-R289 but we lack the test about the document... I'll add

ysugimoto:
@wheresrhys Through the reflect your review, I changed `getParser()` argument name to be more meaningful one:

```
const getParser = (
	schema,
	{
		type,
		titleNodeName = 'name',
		descriptionNodeName = 'description',
		blacklistPropertyNames = [],
	}
)
```

packages/tc-markdown-parser/lib/markdown-parser.js
```diff
@@ -0,0 +1,97 @@
...
+			})
+			.use(stringifyBoast, {
+				titleNodeName,
+				descriptionNodeName,
+			});
+	};
+};
+
+const getParser = (
+	schema,
```
wheresrhys: don't require schema as parameter, or pass around as a variable. Just require tc-schema-sdk at the top of this file and others that need it

PR #95, opened 2019-12-04
V0.1

wheresrhys:
This fixes all the bugs in treecreeper needed to power biz-ops-api

Main changes
- add deprecated fields to change logs
- fix which logs are sent by absorb
- fix sending events for changes to documents
- a few other bug fixes


packages/tc-api-rest-handlers/lib/events/__tests__/docstore-events.spec.js
```diff
@@ -0,0 +1,355 @@
...
+const { setupMocks } = require('../../../../../test-helpers');
+const { postHandler, patchHandler, absorbHandler } = require('../../..');
+
+const eventTester = desiredAction => (type, code, updatedProperties) => {
+	const matchingEvents = emitSpy.mock.calls.filter(
+		([action, event]) =>
+			action === desiredAction &&
+			event.action === desiredAction &&
+			event.type === type &&
+			event.code === code,
```
ysugimoto: Nit: we can use `emitSpy.mock.calls.find(...)` instead of filter then we can avoid continuous array indexing at assertion of `expect(matchingEvents[1]).toMatchObject(referenceObject);`

wheresrhys:
sister PR is here https://github.com/Financial-Times/biz-ops-api/pull/407

packages/tc-api-rest-handlers/absorb.js
```diff
@@ -220,12 +242,42 @@ const absorbHandler = ({ documentStore } = {}) => async input => {
...
+				addedRelationships,
+				removedRelationships,
+				updatedProperties: [
+					...Object.keys(writeProperties),
+					...Object.keys(addedRelationships),
+					...(updatedDocumentProperties || []),
+				],
+				neo4jResult: result,
+			},
+		]);
```
ysugimoto: Could we make these events via some event creation function which is exported by events.js with required arguments e.g action, code and type, and other optional object argument?  This is because making event parameters has pretty much kinds, Is it worth to do?

packages/tc-api-rest-handlers/lib/events/__tests__/docstore-events.spec.js
wheresrhys: I did have that, but that is vulnerable to sending multiple events when only one is intended

packages/tc-api-rest-handlers/absorb.js
wheresrhys: It might be. More or less that is already what's happening internally - `broadcast()` calls the `makeEvents()` function. But for all use cases apart from absorb there is just a single call, so we would end up replacing `broadcast(...)` with `broadcast(makeEvents(...))` in most handlers.

ysugimoto: Ah! I agree with you.

PR #101, opened 2019-12-09
Validate relationship properties

fenglish:
POST and PATCH should use schema for validation and accept payload as defined in [trello ticket 157](https://trello.com/c/ApJqFYv4/157-implement-returning-properties-on-relationships-in-graphql)


fenglish:
@wheresrhys Thank you for your suggestions. I will reflect them by another PR 👍

PR #103, opened 2019-12-09
Parse nested object

ysugimoto:
## Why

Now Treecreeper's markdown parser supports rich relationship property definition, so we should support to define in the markdown i.e runbook.md

## What

This PR improves markdown parser with supporting multiple relationship definition like:

```
## younger siblings

- ChildType
  propNameOne: propValueOne
  propNameTwo: propNameTwo
  ...
```

Note that this PR also includes breaking change which nested definition -- mainly relationship will return an object, not a string.

### before

`hasMany: false` case:

```
## favourite child

some child

//->  favouriteChild: 'some child'
```

`hasMany: true` case:

```
## children

* some child

//->  children: ['some child']
```

### after

`hasMany: false` case:

```
## favourite child

some child

//->  favouriteChild: { code: 'some child' }
```

`hasMany: true` case:

```
## children

* some child

//->  children: [{ code: 'some child' }]
```

## nested property validation

I defined a spec that nested property name should be a lower case, so

- some string: value -> NG
- someString: value -> OK

Of course, this rule can be looser, which accepts white space character, etc...

Trello: https://trello.com/c/KkCSNp47/158-implementing-writing-properties-on-relationships-in-runbookmd


packages/tc-markdown-parser/__tests__/property-value.spec.js
```diff
@@ -334,3 +348,199 @@ describe('dealing with html comments', () => {
...
+
+				example-code
+					some string: i like it
+			`);
+
+			expect(errors).toHaveLength(1);
+			expect(data.name).toEqual('name');
+			const [{ message }] = errors;
+			expect(message).toMatch(/should be lower camel case/);
+			expect(message).toMatch(/line 6/);
```
ysugimoto: It was hard to overwrite `line` property in unified AST, so I intended to include the actual line number inside the error message

wheresrhys:
@chee Woudl be great if you coudl review this - we are enhancing runbook.md to
- turn into a slightly more generic parser of md into biz ops (done in a previous PR)
- add the ability to nest extra key value pairs within lists

packages/tc-markdown-parser/lib/property-coercers/index.js
```diff
@@ -72,6 +73,67 @@ module.exports = {
...
+		if (hasMany && items.length) {
+			try {
+				const values = items.map(parseMultilineDefinition);
+				return {
+					valid: true,
+					value: values,
+				};
+			} catch (error) {
+				return {
+					valud: false,
```
wheresrhys:
```suggestion
					valid: false,
```

packages/tc-markdown-parser/lib/parse-multiline-definition.js
```diff
@@ -0,0 +1,176 @@
...
+ - child-code
+   propNameOne: propValueOne
+
+ Divides to:
+ {
+   code: 'child-code',
+   propDef: '  propNameOne: propValueOne\n',
+ }
+*/
+const dividePropertyDefinition = subdocument => {
```
wheresrhys: Are there tools built into mdast that can do this without us having to manually slice it up @chee ?

packages/tc-markdown-parser/lib/parse-multiline-definition.js
```diff
@@ -0,0 +1,176 @@
...
+		const hasChildren = checkNodeHasChildren(node);
+
+		if (hasTextValue) {
+			const {
+				value,
+				position: {
+					start: { line },
+				},
+			} = node;
+			const { code, propDef } = dividePropertyDefinition(value);
```
wheresrhys: propDefs would be clearer?

packages/tc-markdown-parser/lib/tree-mutators/coerce-treecreeper-properties-to-type.js
```diff
@@ -36,8 +36,10 @@ function getCoercer({ isNested, primitiveType, propertyType }) {
 	const isSubdocument = subdocumentPropertyTypes.has(propertyType);
 	const isString = isNested || (primitiveType === 'String' && !isSubdocument);
 
-	if (isNested || isString) {
-		return propertyCoercers.String;
+	if (isString) {
+		return isNested
+			? propertyCoercers.NestedString
+			: propertyCoercers.String;
```
chee: Nice ^_^

packages/tc-markdown-parser/lib/parse-multiline-definition.js
chee: thanks for inviting me to this pull request, this is all such nice code. and so wonderfully commented!!

i don't think there is something like this, i don't think `remark` even has a plugin for kramdown-style definition lists.

packages/tc-markdown-parser/__tests__/property-value.spec.js
```diff
@@ -334,3 +348,199 @@ describe('dealing with html comments', () => {
...
+
+	describe('thorws syntax error on mutiline defitniion', () => {
+		it('mutiline property name should be a lower camel case', async () => {
+			const { data, errors } = await parser.parseMarkdownString(here`
+				# name
+
+				## favourite child
+
+				example-code
+					some string: i like it
```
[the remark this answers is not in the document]
ysugimoto: Yes, this is test for the error case. And now we're checking property name with regular expression at here https://github.com/Financial-Times/treecreeper/pull/103/files/87789563c0cd83e3e437601b42a467aaafc56f58#diff-ab7fbc57659cb853a1fbf6f572881d6fR74-R78

packages/tc-markdown-parser/__tests__/property-value.spec.js
```diff
@@ -334,3 +348,199 @@ describe('dealing with html comments', () => {
...
+			`);
+
+			expect(errors).toHaveLength(1);
+			expect(data.name).toEqual('name');
+			const [{ message }] = errors;
+			expect(message).toMatch(/unexpected linefeed token found/);
+			expect(message).toMatch(/line 6/);
+		});
+
+		it('property value must not be empty', async () => {
```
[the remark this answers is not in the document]
ysugimoto: Right, but I forgot to consider about other data types. I'll add that

packages/tc-markdown-parser/lib/parse-multiline-definition.js
```diff
@@ -0,0 +1,176 @@
+class SyntaxError extends Error {}
+
+/*
+ This function divides from mutipile definition to code and property string e.g,
```
JakeChampion:
```suggestion
 This function divides from multiple definition to code and property string e.g,
```

packages/tc-markdown-parser/lib/parse-multiline-definition.js
```diff
@@ -0,0 +1,176 @@
...
+};
+
+// Token characters of meaningful property descriptor
+const TAB = '\t';
+const PROPERTY_NAME_SEPARATOR = ':';
+const LINE_FEED = '\n';
+const CARRIAGE_RETURN = '\r';
+
+/*
+ This is tiny lexer/parser fuction for analyze multiple property definition.
```
JakeChampion:
```suggestion
 This is tiny lexer/parser function for analyze multiple property definition.
```

packages/tc-markdown-parser/lib/parse-multiline-definition.js
```diff
@@ -0,0 +1,176 @@
...
+   }
+*/
+const parsePropertyDefinition = (propDefString, line) => {
+	const properties = {};
+	const strlen = propDefString.length;
+	let buffer = '';
+	let propName = '';
+	let column = 0;
+
+	for (let i = 0; i < strlen; i++) {
```
JakeChampion: This could be replaced with a `for (const char of propDefString) {`, do you think that would be a good change to make?

ysugimoto: With thinking of loop performance, I chose primitive for-loop but I didn't realize ES6 syntax for-of loop. it looks neat! I'll use this, thanks!

ysugimoto:
I changed the approach to deal with parsing multi-line definition:

## Not to do in coercing phase, do in unified transforming phase

I added more tree-mutator, `setNestedMultilineProperties`, which analyze AST and transform to expected property nodes for nested property definition.
To being transformed by `setNestedMultilineProperties`, it has the additional guarantee that all nested property can be dealt with. Therefore, the `property-coercer`can keep it simple -- especially String coercer doesn't need to consider `hasMany` or not anymore. 

In addition, I included the fix of https://trello.com/c/njf30klB/220-error-if-description-is-in-multiple-blocks. If the user wrote descriptions in multiple blocks, then we'll raise an error

packages/tc-markdown-parser/__tests__/first-paragraph.spec.js
```diff
@@ -53,3 +53,27 @@ test('top level content in an h2-range is not parsed as description', async () =
...
+	const [{ message }] = errors;
+	expect(message).toMatch(
+		/only one description is allowed to define in top level/,
+	);
+	expect(data).toMatchObject({
+		name: 'well',
+		description: 'this is first description',
+		someString: "how's tricks",
+	});
+});
```
ysugimoto: This is test for https://trello.com/c/njf30klB/220-error-if-description-is-in-multiple-blocks

packages/tc-markdown-parser/lib/property-coercers/index.js
```diff
@@ -37,40 +37,43 @@ module.exports = {
...
+				value: 'expected a single item, but got a bulleted list',
 			};
 		}
 
-		// Not expecting a list, didn't get a list'
-		if (!hasMany && !items.length) {
+		return {
+			valid: true,
+			value: flattenNodeToPlainString(subdocument),
+		};
```
ysugimoto: No longer we don't need to consider `hasMany` in string coercer because nested definition will be processed at NestedString coercer

packages/tc-markdown-parser/lib/tree-mutators/create-treecreeper-description-node.js
```diff
@@ -16,6 +17,13 @@ module.exports = function createBizopsDescriptionNode({
 			// if we have no children there's no point in carrying on
 			return;
 		}
+		if (descriptionChildren.length > 1) {
+			convertNodeToProblem({
+				node: descriptionChildren.pop(),
+				message:
+					'only one description is allowed to define in top level',
+			});
+		}
```
ysugimoto: validates desciprion node exists multiply

packages/tc-markdown-parser/lib/tree-mutators/set-nested-multiline-properties.js
```diff
@@ -0,0 +1,84 @@
...
+			if (error.position) {
+				node.position = error.position;
+			}
+		}
+	}
+
+	return function transform(tree) {
+		visit(tree, 'property', mutate);
+	};
+};
```
ysugimoto: All nested property definitions are processed in this tree-murator

PR #105, opened 2019-12-10
Preact browser

wheresrhys:
# Why
In order to  build a UI for rich relationships we need to first move the existing UI into treecreeper and make it more easy to iterate on and test. An isomorphic preact component is what I've gone with

# What
- creates a `tc-ui` package, which (only implemented for relationship so far), for each type of component exports 
```
{
  EditComponent, // for rendering an input in the edit form
  ViewComponent, // for rendering the data on the view page
  parser // function for converting data submitted by the form into a property to be used by the REST PATCH request
}
```
- Also export a `browser.jsx` file, which has a method `attachEditComponent` for attaching an interactive component to a container on the client side
- Created an interactive, isomorphic `relationship-picker.jsx` component. 

One vast improvement in all this is that the data collected by relationship-picker is stored as json in a hidden input - much neater than the several inputs that needed parsing before, and allowing for a lot of simplification in the server side data parsing. Handling client side errors, and preventing form submission, is also a little neater, but will revisit again when more components are implemented.

I also fixed a bug whereby writing a relationship with no codes e.g. `youngerSiblings: []` was throwing an error.

### TODO
(in next PR)
- [ ] Add test for submitting empty array for relationships
- [ ] Add lots of tests for the relationship component


packages/tc-ui/relationship/browser.jsx
```diff
@@ -0,0 +1,11 @@
+const { render, h } = require('preact');
+const { RelationshipPicker } = require('./lib/relationship-picker.jsx');
+
+module.exports = {
+	attachEditComponent: container =>
+		render(
+			<RelationshipPicker {...JSON.parse(container.dataset.props)} />,
+			document,
+			container,
+		),
```
ysugimoto: naming: this looks like HoC component, so the name should be `withEditComponent`, what do you think?

webpack.config.js
```diff
@@ -10,6 +10,14 @@ module.exports = {
...
 	resolve: {
 		extensions: ['.js', '.jsx', '.css'],
+		alias: {
+			react: 'preact/compat',
+			'react-dom': 'preact/compat',
+			// Not necessary unless you consume a module using `createClass`
+			'create-react-class': 'preact/compat/lib/create-react-class',
+			// Not necessary unless you consume a module requiring `react-dom-factories`
+			'react-dom-factories': 'preact/compat/lib/react-dom-factories',
+		},
```
ysugimoto: 👍 

I remember once biz-ops-admin UI had a bundling JS file issue that transpile via babel won't raise error although missing component and some cases, so installing webpack will solve that concern.

packages/tc-ui/relationship/lib/relationship-picker.jsx
```diff
@@ -0,0 +1,329 @@
...
+			searchTerm: '',
+			isUserError: false,
+			isUnresolved: false,
+		});
+		// this is needed to prevent the event propagating up  and then
+		// immediately clicking another button. VERY odd behaviour, and
+		// don't fully understand why, but this is the fix
+		if (event) {
+			event.preventDefault();
+		}
```
ysugimoto: I also investigated that. I'm guessing that `ReactAuthSuggest` won't care about event propagation, but I'm not sure. Anyway, this is good to fix

packages/tc-ui/relationship/browser.jsx
wheresrhys: what does HoC mean?

wheresrhys: Ah yes - I read about it. I'll change this in the PR I'm working on now (which is based off this branch) in order to avoid having to rebase.

PR #107, opened 2019-12-13
Wip support list definition

ysugimoto:
This PR supports some list definitions of string and enums in relationship properties (currently not in a top-level).

### Spec

I specified list definition as Yaml-like syntax:

```
## curious child

example-code:
  someListString:
    - foo
    - bar
  someListEnums:
    - First
    - Second
```

At the moment, parse seems ok but I assume my implementation looks fragile and it should be more robustly.

## Does not include

- nested property validation in `tc-schema-sdk`, so we need to implement them in here: https://github.com/Financial-Times/treecreeper/blob/master/packages/tc-schema-sdk/lib/validators.js#L73-L76
- List type schema definition, but it will be satisfied via https://github.com/Financial-Times/treecreeper/pull/67

I'm sorry for leaving it as draft PR, but feel free to destroy them.


wheresrhys:
Replaced by this https://github.com/Financial-Times/treecreeper/pull/147 if you want to compare @ysugimoto

PR #130, opened 2020-01-06
Fix relationship removal

wheresrhys:
# Why
Relationship removal is broken https://financialtimes.slack.com/archives/C07B3043U/p1578309991018800

# What
Fixes missing target property on an event. Comment in the code explains what I _think_ is going on


wheresrhys:
I am gonna go ahead and merge this as it is __*broken in prod*__

PR #132, opened 2020-01-15
setup e2e test with Cypress

kiyaGu:


cypress/integration/end-to-end-journey_spec.js
```diff
@@ -0,0 +1,52 @@
...
+				input.val('2020-01-15T13:00');
+			})
+			.click();
+		cy.get('input[name=someUrl]').type('https://google.com');
+
+		cy.get('[data-testid="submit"]').click();
+	});
+
+	it('Navigates through an end-to-end journey for creating MainType', () => {
+		cy.visit(`/MainType/e2e-demo`);
```
wheresrhys: add checks for field values too

package.json
```diff
@@ -58,7 +60,11 @@
   },
   "scripts": {
     "test": "make test",
-    "postinstall": "athloi exec npm i -- --no-package-lock"
+    "postinstall": "athloi exec npm i -- --no-package-lock",
+    "cy:verify": "cypress verify",
```
wheresrhys: put these tasks in make

test-helpers/db-connection.js
```diff
@@ -1,7 +1,7 @@
 const neo4j = require('neo4j-driver').v1;
 
 const driver = neo4j.driver(
-	process.env.NEO4J_BOLT_URL,
+	process.env.NEO4J_BOLT_URL || 'bolt://localhost:7687',
```
wheresrhys: once the tasks are in make, you will get the env vars for free and not need this

packages/tc-ui/src/lib/components/buttons.jsx
```diff
@@ -7,6 +7,7 @@ const getButtonClasses = extraClasses =>
 
 const EditButton = props => (
 	<a
+		data-testid="edit"
```
wheresrhys: is there some way we can hook onto existing data-attributes, or if we need data attributes/classes to pick out these buttons in tests, just use non test names e.g. `data-button-type="edit"`

demo/cms/render/react-renderer.jsx
```diff
@@ -12,7 +12,7 @@ const assetManifest = {
 	'main.js': 'main.js',
 };
 
-const ASSET_ROOT = 'http://local.in.ft.com:8080/statics/';
+const ASSET_ROOT = '/';
```
wheresrhys: does this still work when running the demo app locally and outside of cypress i.e. `make run`?

wheresrhys: do we even need webpack dev server any more? I'm happy to remove it

.circleci/config.yml
```diff
@@ -114,6 +174,10 @@ workflows:
       - install:
           context: rel-eng-creds
           filters: *only_version_tags
+      - build:
+          context: rel-eng-creds
```
wheresrhys: shouldn't need this line

.circleci/config.yml
```diff
@@ -123,11 +187,17 @@ workflows:
           requires:
             - install
           filters: *only_version_tags
+      - end-to-end-test:
+          context: rel-eng-creds
```
wheresrhys: shouldn't need this line

.circleci/config.yml
```diff
@@ -88,6 +121,33 @@ jobs:
...
+    <<: *default_container_config
+
+    steps:
+      - *attach_workspace
+
+      - run: sudo sh -c 'echo "127.0.0.1 local.in.ft.com" >> /etc/hosts'
+      - run:
+          name: Prepare neo4j database
+          command: |
+            java -version
```
wheresrhys: see if you can put this setup script in a make task shared between the 2 jobs that use it now

demo/cms/render/react-renderer.jsx
kiyaGu: yes, it does work

test-helpers/db-connection.js
kiyaGu: this is not working locally despite putting the tasks in make ... 
```
Uncaught TypeError: Bolt URL expected to be string but was: undefined
```

cypress/integration/e2e-create-child-journey_spec.js
```diff
@@ -0,0 +1,23 @@
+const { code } = require('../fixtures/mainTypeData.json');
```
wheresrhys: do we still need this test now? As long as we can test creation of a record of some type (which we do for MainType) there's not much additional value in testing for other types?

cypress/integration/e2e-create-parent-journey_spec.js
```diff
@@ -0,0 +1,23 @@
+const { code } = require('../fixtures/mainTypeData.json');
+const { populateParentTypeFields, save } = require('../test-helpers');
+const { dropFixtures } = require('../../test-helpers/test-fixtures');
+
+const resetDb = async () => {
+	await dropFixtures(code);
+};
+
+describe('End-to-end journey for creating ParentType', () => {
```
wheresrhys: can also remove this test file

cypress/integration/e2e-create-main-journey_spec.js
```diff
@@ -0,0 +1,53 @@
...
+	populateMainTypeFields,
+	visitMainTypePage,
+	save,
+} = require('../test-helpers');
+
+const resetDb = async () => {
+	await dropFixtures(code);
+};
+
+describe('End-to-end Journey', () => {
```
wheresrhys:
```suggestion
describe('End-to-end Journey - record creation', () => {
```

test-helpers/db-connection.js
wheresrhys: err, ok, leave it as it is then

cypress/integration/e2e-create-child-journey_spec.js
kiyaGu: true we don't need them - they are also covered as part of relationship creation

PR #135, opened 2020-01-21
upgraded to use our neo4j-graphql-js fork

wheresrhys:


packages/tc-ui/src/pages/view/template.jsx
```diff
@@ -19,7 +19,7 @@ const Properties = ({ fields, data, assignComponent }) => {
 		)
 		.map(([name, item]) => {
 			const viewModel = {
-				value: data[name],
+				value: data[name] || data[`${name}_rel`],
```
wheresrhys: This will be vulnerable to e.g. if the value is `false`, then it will end up looking for `_rel` and displaying undefined

A more robust approach - and easier to read the intent I think, would be `\`${name}_rel\` in data ?  data[`${name}_rel`] : data[name]`

Even that's not perfect as the check we really want is 'is `name` the name of a relationship property, in that case use _rel, otherwise just use name', so a condition which checks for the relationshipness of `name` would be ideal

packages/tc-ui/src/primitives/relationship/lib/relationship-picker.jsx
```diff
@@ -234,6 +234,7 @@ class RelationshipPicker extends React.Component {
 							disabled={disabled}
 							onRelationshipRemove={this.onRelationshipRemove}
 							index={i}
+							key={i}
```
wheresrhys: this is for linting?

packages/tc-ui/src/primitives/relationship/lib/view-relationship.jsx
```diff
@@ -1,15 +1,16 @@
 const React = require('react');
+const { sortBy } = require('lodash');
```
wheresrhys: this will be used on the client side. To make the code bundel smaller we can instead `npm install lodash.sortby` then `sortBy = require('lodash.sortby');`

packages/tc-ui/src/lib/components/structure.jsx
```diff
@@ -128,7 +128,7 @@ const MetaProperties = ({ data, isCreate }) => {
 
 	return (
 		<>
-			{isCreate ? 'Created by ' : 'Last updated by '}
+			{isCreate ? 'Created by ' : ' Last updated by '}
```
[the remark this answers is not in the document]
kiyaGu: I have added this to add whitespace 
Screenshot 2020-01-29 at 15 16 25

packages/tc-ui/src/primitives/relationship/lib/relationship-picker.jsx
kiyaGu: there was a React warning ```Each child in an array should have a unique "key" prop.
Check the render method ...```

packages/tc-ui/src/pages/edit/template.jsx
```diff
@@ -15,16 +16,24 @@ const getValue = (itemSchema, itemValue) => {
 	if (itemSchema.relationship) {
 		if (itemSchema.hasMany) {
 			return itemValue
-				? itemValue.map(item => ({
-						code: item.code,
-						name: item.name || item.code,
+				? sortBy(itemValue, `${itemSchema.type}.code`).map(item => ({
```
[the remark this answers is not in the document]
kiyaGu: @wheresrhys I have moved the entire getValue function to Primitives/server.js or were you asking to move only the relationship part to relationship component?

packages/tc-ui/src/lib/components/structure.jsx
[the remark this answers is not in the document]
wheresrhys: oh ok

packages/tc-ui/src/pages/view/template.jsx
```diff
@@ -18,13 +18,16 @@ const Properties = ({ fields, data, assignComponent }) => {
 				!schema.deprecationReason,
 		)
 		.map(([name, item]) => {
+			const value = item.isRelationship
```
wheresrhys: :+1:

packages/tc-ui/src/pages/edit/template.jsx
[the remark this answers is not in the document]
wheresrhys: let's talk about it a bit more tomorrow, but it's already a step in the right direction

PR #140, opened 2020-01-30
support multiple choice enums in markdown

wheresrhys:
# Why
We can't support any format of data we want to use in runbooks in the API and UI without also supporting it in the markdown parser. We recently added support for multiple choice enums to the API and UI, so here they are in markdown

# What
Adds the ability for the enum coercer to parse items in a list

## todo
- [ ] add support for multiple choice as properties on relationships

### DIFF
--- a/packages/tc-markdown-parser/lib/tree-mutators/coerce-treecreeper-properties-to-type.js
+++ b/packages/tc-markdown-parser/lib/tree-mutators/coerce-treecreeper-properties-to-type.js
@@ -109,7 +109,10 @@ const coerceNestedPropertyValue = (
 	}
 };
 
-const coerceEnumPropertyValue = (node, { propertyType: enumName, hasMany, enums }) => {
+const coerceEnumPropertyValue = (
+	node,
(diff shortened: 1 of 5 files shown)


packages/tc-markdown-parser/lib/tree-mutators/coerce-treecreeper-properties-to-type.js
```diff
@@ -109,23 +109,51 @@ const coerceNestedPropertyValue = (
...
+		}
+		[subdocument] = subdocument.children;
+	} else if (subdocument.children[0].type !== 'paragraph') {
+		return convertNodeToProblem({
+			node,
+			message: 'Must provide a single enum, not a nested list',
+		});
+	}
 
 	const enumName = propertyType;
```
kiyaGu: can this be moved to 
```const coerceEnumPropertyValue = (node, { propertyType: enumName, hasMany, enums }) => {...```

chee:
:) looks great

PR #142, opened 2020-02-04
Move tests

kiyaGu:
- move tests so that they live in the same directory as the component they are testing


packages/tc-ui/src/pages/view/__tests__/e2e-view-complete-record.cyp.js
```diff
@@ -8,17 +8,13 @@ const {
 	anotherString,
 	someUrl,
 	promptText,
-} = require('../fixtures/mainTypeData.json');
-const { dropFixtures } = require('../../test-helpers/test-fixtures');
+} = require('../../../../../../cypress/fixtures/mainTypeData.json');
```
wheresrhys: could these cypress fixtures move inside tc-ui too? Long chains of `../` can make refactoring painful

PR #147, opened 2020-02-05
Relationship multienum

wheresrhys:
# Why
We need to support editing multiple choice fields when they annotate relationships in markdown files. 

# What

The authoring experience is as follows:
```md
## property name

i-am-a-code
  someMultipleChoice: first,second
```
Previously, the parser did not carry out any enum validation or coercing for nested fields. The parser now works by converting the nested field `someMultipleChoice: first,second` into an AST identical to if the input were:
```
## some multiple choice
- first
- second
```
Then the property coercer is able to run the top level coercers on the nested property list. The same code path is followed for all properties on relationships, which I think makes a lot of sense. I have found very few simplifications of code elsewhere as a result though, one exception being a few less conditions in coerceTreecreeperPropertiesToType

Unfortunately a bit hard to review as I've had to move some code blocks around due to linting rules about using before declaring functions


packages/tc-markdown-parser/lib/parse-multiline-definition.js
```diff
@@ -17,6 +17,10 @@ const createNode = (key, value, line, column) =>
 		key,
 		position: {
 			start: {
+				line,
+				column: 0,
+			},
+			end: {
```
wheresrhys: the value for column is taken when the parser _finishes_ parsing a nested property, so I think it was a bug before to use it as `start`

packages/tc-markdown-parser/__tests__/nested-property-value.spec.js
```diff
@@ -371,7 +438,7 @@ describe('nested property definition tests', () => {
 			});
 		});
 
-		it('can be parsed as object array which contains property object', async () => {
+		it('can be parsed as Array of object which additional property definitions', async () => {
```
kiyaGu: is **which** a typo for with?

PR #148, opened 2020-02-06
Display rich-relationship properties

kiyaGu:
Screenshot 2020-02-06 at 09 43 25

### DIFF
--- a/packages/tc-ui/src/primitives/relationship/__tests__/e2e-create-relationships.cyp.js
+++ b/packages/tc-ui/src/primitives/relationship/__tests__/e2e-create-relationships.cyp.js
@@ -37,19 +37,22 @@ describe('End-to-end - relationship creation', () => {
 				.should('have.attr', 'href', `/ChildType/${code}-first-child`);
 		});
 
-		it('can not select another selection for one-to-one relationship', async () => {
+		it('can not select another selection for one-to-one relationship', () => {
 			visitMainTypePage();
 			visitEditPage();
 
 			pickFavouriteChild();
+			cy.get('#ul-favouriteChild span').should(
+				'have.text',
+				'e2e-demo-first-child',
+			);
+			cy.get('#favouriteChild-picker').should('be.disabled');
 			save();
 
 			cy.url().should('contain', `/MainType/${code}`);
 			cy.get('#favouriteChild')
 				.should('have.text', `${code}-first-child`)
 				.should('have.attr', 'href', `/ChildType/${code}-first-child`);
-
-			cy.get('#favouriteChild-picker').should('be.disabled');
 		});
 	});
 
@@ -149,7 +152,7 @@ describe('End-to-end - relationship creation', () => {
 			.should('have.attr', 'href', `/ParentType/${code}-parent`);
 	});
 
-	it('fetches suggestions from the autocomplete api', async () => {
+	it('fetches suggestions from the autocomplete api', () => {
 		const selector = '[data-property-name="children"]';
 		visitMainTypePage();
 		visitEditPage();
@@ -184,7 +187,7 @@ describe('End-to-end - relationship creation', () => {
 		);
 	});
 
-	it('does not suggest previously selected records', async () => {
+	it('does not suggest previously selected records', () => {
 		const selector = '[data-property-name="children"]';
 		visitMainTypePage();
 		visitEditPage();
--- a/packages/tc-ui/src/primitives/relationship/__tests__/e2e-create-relationships.cyp.js
+++ b/packages/tc-ui/src/primitives/relationship/__tests__/e2e-create-relationships.cyp.js
@@ -215,7 +215,6 @@ describe('End-to-end - relationship creation', () => {
 	});
 
 	it('can create main -> parent relationship', () => {
-
 		pickParent();
 		save();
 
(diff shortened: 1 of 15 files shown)


packages/tc-ui/src/primitives/relationship/__tests__/e2e-create-relationships.cyp.js
```diff
@@ -23,6 +23,121 @@ describe('End-to-end - relationship creation', () => {
...
+			pickFavouriteChild();
+			save();
+
+			cy.url().should('contain', `/MainType/${code}`);
+			cy.get('#favouriteChild')
+				.should('have.text', `${code}-first-child`)
+				.should('have.attr', 'href', `/ChildType/${code}-first-child`);
+		});
+
+		it('can not select another selection for one-to-one relationship', () => {
```
wheresrhys: I think there are 3 things to test here
- is it disabled on page load if relationship already exists
- does it get undisabled if relationship is removed
- does it get re-enabled if relationship is added again

Not sure how to structure the tests, but these are all important cases

packages/tc-ui/src/primitives/relationship/__tests__/e2e-create-relationships.cyp.js
```diff
@@ -23,6 +23,121 @@ describe('End-to-end - relationship creation', () => {
...
+					.find('a')
+					.should(
+						'have.attr',
+						'href',
+						`/ChildType/${code}-second-child`,
+					);
+			});
+		});
+
+		it('can select another selection for one-to-many relationship', () => {
```
wheresrhys:
```suggestion
		it('does not disable when relationship already exists', () => {
```
This test shoudl move before the previous one too I think

packages/tc-ui/src/primitives/relationship/__tests__/e2e-create-relationships.cyp.js
```diff
@@ -23,6 +23,121 @@ describe('End-to-end - relationship creation', () => {
...
+				.should('have.attr', 'href', `/ChildType/${code}-first-child`);
+
+			cy.get('#children>li')
+				.eq(1)
+				.should('have.text', `${code}-second-child`)
+				.find('a')
+				.should('have.attr', 'href', `/ChildType/${code}-second-child`);
+		});
+
+		it('can select a selection for one-to-many relationship', () => {
```
wheresrhys:
```suggestion
		it('can select multiple selections for one-to-many relationship', () => {
```

packages/tc-ui/src/primitives/relationship/lib/view-relationship.jsx
```diff
@@ -1,25 +1,76 @@
...
+	LabelledPrimitive,
+} = require('../../../lib/components/structure');
 
 let RelationshipAnnotator;
 
-const OneRelationship = ({ type, value = {}, id }) =>
-	type && value[type] ? (
+const OneRelationship = props => {
+	const {
+		assignComponent,
```
wheresrhys: you mention in a comment before that you'd like to find a better way to pass assignComponent around. I agree that this is a little ugly, but can't think of a better way either. 

@ericanand do you think React hooks might help us be able to pass a reference to a utility function around?

packages/tc-ui/src/primitives/relationship/server.jsx
```diff
@@ -54,6 +54,11 @@ module.exports = {
 		Object.entries(typeDef.properties)
 			.filter(([, { useInSummary }]) => useInSummary)
 			.forEach(([name]) => props.add(name));
-		return `${propName}_rel {${type} {${[...props].join(' ')}}}`;
+		const relationshipProps = [...new Set(Object.keys(properties))].join(
+			' ',
+		);
+		return `${propName}_rel {${type} {${[...props].join(
```
wheresrhys: assign `[...props].join(' ')` to a variable on the previous line so that we avoid prettier formatting like this? It'd be nice to have it returning `${propName}_rel {${type} {${nodeProps}} ${relationshipProps}}` all on one line (if it fits) as it's more readable

packages/tc-ui/src/primitives/relationship/lib/view-relationship.jsx
ericanand: I probably need more context about this, but if you're finding your passing something around alot, maybe [contexts](https://reactjs.org/docs/context.html) is what you want? (I feel I missed out on a joke here with me lacking context about this but oh well)

Happy to chat about this.

kiyaGu: Will have a look into that

packages/tc-ui/src/primitives/relationship/lib/view-relationship.jsx
```diff
@@ -85,13 +85,12 @@ const ViewRelationship = props => {
...
 		<ul id={id} className="o-layout__unstyled-element biz-ops-links">
 			{sortBy(value, [`${type}.code`]).map((item, index) => {
-				const modifiedProps = { ...props, value: item };
 				return (
 					<li
 						key={index}
 						className={inactiveCheck(item) ? 'inactive' : 'active'}
 					>
-						<OneRelationship {...modifiedProps} />
+						<OneRelationship {...{ ...props, value: item }} />
```
wheresrhys: can this just be ?
```suggestion
						<OneRelationship { ...props, value: item } />
```

kiyaGu: this will not work `Unexpected token, expected "}"`

PR #149, opened 2020-02-12
Fix markdown in description

wheresrhys:
# Why
In some runbook.md, notably https://github.com/Financial-Times/monitoring-aggregation-ecs/blob/master/documentation/RUNBOOK.md, the description contains markdown. When this is stringified as a subdocument additional newlines are added, causing the content that is saved to Biz Ops to be quite distorted from the original intent

# What
~So far only adds a test to recreate the problem. Needs more work to fix the subdocument rendering~

The problem lies with the fact that teh rendering of the description AST back to markdown uses the same function - renderSubdocument - that's used for other purposes. This function usually works with converting single fields that are expected to contain lots of markdown back into markdown so, as some sort of precaution, converts the top level node to a `root` node. However, for description we need the top level node to be a `p`, otherwise the inlined stuff gets uninlined. So the first step of fixing this was to allow passing a flag to decide whether to coerce the tree to be a markdown `root` node or not.

This ended up breaking a load of tests for coercing the h1 field. These were side-effects of the fact that the tests for error states in h1 parsing were inadvertently having the h1 converted to a problem node, which the parser then attempted to convert to a description... which never resulted in a final error because the trick above converted the problem node into a `root` node before trying to apply remark-stringify. There was also a similar case where, if using a custom choice of field to convert the h1 to, this wasn't picked up on by the filter looking for the 'next content after the heading', but this was also failing silently.

Anyway, that's all fixed now as choice of title field is passed into and used correctly by the description coercer.

Other than that, this PR is mainly linting

*Note that this is forked from the last 0.3.9 release. Once approved, i will publish again from this branch, use in the RUNBOOK.md app, then merge to master here*


packages/tc-markdown-parser/lib/property-coercers/index.js
```diff
@@ -83,7 +83,10 @@ module.exports = {
 		if (hasMany) {
 			return {
 				valid: true,
-				value: split(subdocument).map(renderSubdocument),
+				// avod passing index as second parameter
```
kiyaGu: typo - `avoid`

wheresrhys:
```suggestion
				// avoid passing index as second parameter
```

PR #150, opened 2020-02-14
Edit rich relationship

kiyaGu:
This is a working version (while ignoring errors like ```Warning: Encountered two children with the same key, `2`. Keys should be unique so that components maintain their identity across updates``` and the consequence of these errors when updating records)
TODO
- fix ☝️ (the error is fixed but still get error on saving and this will be fixed in the next ticket)


packages/tc-ui/src/lib/components/input-wrapper.jsx
```diff
@@ -50,7 +54,7 @@ const FieldTitle = ({ label, description, expandableContent, lockedBy }) => (
 const WrappedEditComponent = props => {
 	props = { ...props, disabled: !!props.lockedBy };
 	const { Component } = props;
-	const WrapperTag = props.wrapperTag || 'label';
+	const WrapperTag = props.wrapperTag || 'div';
```
wheresrhys: is this to fix the bug with the target area for the button being the whole row?

packages/tc-ui/src/lib/components/input-wrapper.jsx
```diff
@@ -14,7 +14,11 @@ const FieldTitle = ({ label, description, expandableContent, lockedBy }) => (
 		<span className="o-forms-title__prompt description-text">
 			<span
 				dangerouslySetInnerHTML={{
-					__html: autolinker.link(description),
+					__html:
+						typeof window === 'undefined'
+							? autolinker.link(description)
```
wheresrhys: Interested in why this is necessary. Strange

packages/tc-ui/src/lib/mappers/component-assigner.js
```diff
@@ -0,0 +1,39 @@
...
+	getPrimitiveTypes,
+} = require('@financial-times/tc-schema-sdk');
+
+module.exports = ({ customComponents = {}, customTypeMappings = {} } = {}) => ({
+	type,
+	hasMany,
+}) => {
+	// TO Do - don't require in here
+	// eslint-disable-next-line global-require
+	const primitives = require('../../primitives/server');
```
wheresrhys: why can this not be required at the top level?

packages/tc-ui/src/pages/edit/template.jsx
```diff
@@ -138,6 +137,14 @@ const EditForm = props => {
...
 			</form>
 			<script src="https://cloud.tinymce.com/stable/tinymce.js" defer />
+			<script
+				type="application/json"
+				data-json="schema-data"
+				dangerouslySetInnerHTML={{
+					__html: JSON.stringify({ ...rawData.getAll() }),
+				}}
+			/>
+			<script src="https://cdnjs.cloudflare.com/ajax/libs/autolinker/3.11.1/Autolinker.min.js" />
```
wheresrhys: ah, now I see. If we can, it would be good to understand why autolinker cannot be bundled by webpack successfully

packages/tc-ui/src/primitives/multiple-choice/server.jsx
```diff
@@ -11,7 +11,7 @@ const Checkbox = ({ name, checkboxValue, disabled, checked }) => {
 				value={checkboxValue}
 				aria-label={checkboxValue}
 				id={`checkbox-${name}-${checkboxValue}`}
-				checked={checked}
+				defaultChecked={checked}
```
wheresrhys: is this some react feature?

packages/tc-ui/src/server.js
```diff
@@ -1,6 +1,7 @@
 const primitives = require('./primitives/server');
-
-const { componentAssigner, graphqlQueryBuilder } = require('./lib/mappers');
+const componentAssigner = require('./lib/mappers/component-assigner');
```
wheresrhys: nice idea splitting these

example-schema/relationships/CuriousChild.yaml
```diff
@@ -36,8 +36,3 @@ properties:
     label: Label for some float
     description: Description for some float.
     type: Float
-  someMultipleChoice:
```
wheresrhys: why is this removed?

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "treecreeper",
   "private": true,
   "version": "0.0.0",
-  "description": "A set of tools for building adnmin interfaces on top of graph data",
```
wheresrhys: :joy:

packages/tc-ui/src/lib/mappers/component-assigner.js
```diff
@@ -0,0 +1,45 @@
+const {
```
wheresrhys: Slightly prefer to export `{ componentAssigner }` not `componentAssigner`.

Dora persuaded me that this means there is greater consistency when importing, easier refactoring if you need to add more exports to the same file, and can be easier to write tests for

packages/tc-ui/src/lib/mappers/get-value.js
```diff
@@ -0,0 +1,39 @@
+const sortBy = require('lodash.sortby');
+
+module.exports = (itemSchema, itemValue) => {
```
wheresrhys: similar comment to on previous file, but also add that if you use

```
const myFunc = // define the function

module.exports = { myFunc }

```

then error stack traces and debugger tools will have information about the function name, rather than anonymous function

packages/tc-ui/src/pages/edit/browser.js
```diff
@@ -1,9 +1,17 @@
...
+const { init } = require('@financial-times/tc-schema-sdk');
 const {
 	Relationship: { withEditComponent: attachRelationshipPicker },
 	LargeText: { withEditComponent: attachDocumentEditor },
 } = require('../../primitives/browser');
 
+const initSchema = async () => {
+	const jsonNode = document.querySelector("[data-json='schema-data']");
+	const jsonRaw = jsonNode && jsonNode.textContent;
+	const jsonData = jsonRaw && jsonRaw.length > 4 ? JSON.parse(jsonRaw) : {};
```
wheresrhys: Why checking for `jsonRaw.length > 4` ?

packages/tc-ui/src/primitives/boolean/server.jsx
```diff
@@ -17,7 +17,7 @@ const Checkbox = ({ name, checkboxValue, disabled, userValue }) => {
 				value={checkboxValue.toString()}
 				aria-label={label}
 				id={`radio-${name}-${label}`}
-				checked={userValue === checkboxValue ? 'true' : null}
+				defaultChecked={userValue === checkboxValue ? 'true' : null}
```
wheresrhys: Similar to the select defaultChecked only needing us to pass in the actual user value, ios this the correct thing to do here? I'm not sure
```suggestion
				defaultChecked={userValue}
```

packages/tc-ui/src/primitives/relationship/lib/relationship-picker.jsx
```diff
@@ -78,7 +79,7 @@ class RelationshipPicker extends React.Component {
 		// this is needed to prevent the event propagating up and then
 		// immediately clicking another button. VERY odd behaviour, and
 		// don't fully understand why, but this is the fix
-		event.preventDefault();
+		event.stopPropagation();
```
wheresrhys: ah - was this caused by the `label` thing too?

packages/tc-ui/src/primitives/text/server.jsx
```diff
@@ -1,19 +1,23 @@
...
-			className="o-forms__text"
-			type="text"
-			value={value || null}
-			required={required ? 'required' : null}
-			disabled={disabled}
-		/>
-	</span>
-);
+const EditText = ({ propertyName, value, required, lockedBy, disabled }) => {
+	const name =
```
wheresrhys: I guess you don't need this any more?

webpack.config.js
```diff
@@ -38,6 +38,10 @@ module.exports = {
 			encoding: 'utf8',
 		}),
 	],
+	node: {
+		fs: 'empty',
```
wheresrhys: can you leave a comment explaining why this needs to be here?

packages/tc-ui/src/primitives/relationship/lib/relationship-picker.jsx
kiyaGu: not sure, I this is a more appropriate than preventDefault for this scenario

packages/tc-ui/src/primitives/boolean/server.jsx
kiyaGu: I think this will not work - if you pick 'No' - when you go back to edit button it will not select it as `userValue = false` and if you picked 'Yes` in the edit page it will pick 'No' as your selection as it is the second (both Yes and No have value of defaultSelected = true)

example-schema/relationships/CuriousChild.yaml
kiyaGu: it is defined in line 26

webpack.config.js
kiyaGu: I added that without doing much investigation but it will suppress the following error `Module not found: Error: Can't resolve 'fs'` while I still using `css-loader@3.2.1`
Screenshot 2020-02-14 at 17 46 43
shall I create a ticket to investigate that?

packages/tc-ui/src/pages/edit/browser.js
kiyaGu: no particular reason other than guarding 'null' string

webpack.config.js
kiyaGu: this is resolved now with the refactoring done

example-schema/relationships/CuriousChild.yaml
wheresrhys: oh yes, I remember now

packages/tc-schema-sdk/server.js
```diff
@@ -1,7 +1,6 @@
 const { SDK } = require('./sdk');
 const readYaml = require('./lib/read-yaml');
 
-module.exports = new SDK({ init: false });
+module.exports = new SDK({ init: false, readYaml });
 
 module.exports.SDK = SDK;
```
wheresrhys: if you continue to export readYaml from this file...

packages/tc-schema-validator/tests/enums.js
```diff
@@ -1,7 +1,8 @@
 /* global it, describe, expect */
 const { SDK } = require('@financial-times/tc-schema-sdk');
```
wheresrhys: then you can add `, readYaml` to this first import

packages/tc-schema-validator/tests/enums.js
```diff
@@ -1,7 +1,8 @@
...
+const readYaml = require('@financial-times/tc-schema-sdk/lib/read-yaml');
```
wheresrhys: and remove this import.

The reason this is a good thing to do is because unless a library explicitly documents additional files are meant for requiring, it's safer/less likely to need refactoring to just require the package by name, rather than files internal to the package.

packages/tc-ui/src/primitives/boolean/server.jsx
wheresrhys: ok, I think I follow. Do we have cypress tests for his scenario?

packages/tc-ui/src/primitives/relationship/lib/relationship-picker.jsx
wheresrhys: Have you manually tested that the bad behaviour described in the comment doesn't happen after this change? If so, then :+1:

packages/tc-ui/src/primitives/relationship/lib/relationship.jsx
```diff
@@ -1,25 +1,96 @@
...
+								disabled ? 'disabled' : ''
+							}`}
+							onClick={this.showRichRelationshipEditor}
+							data-index={`annotate-${index}`}
+						>
+							Annotate
+						</button>
+					</span>
+					{isMounted && annotate ? (
+						<RichRelationships key={index} {...this.props} />
```
wheresrhys: Small niggle - can this component be renamed to RichRelationship or RelationshipProperties or something. While the implementation was in flux I wasn't sure what the content of the component would end up being, but now (if I understand correctly) it is a wrapper around the properties of a single relationship

packages/tc-schema-validator/tests/enums.js
kiyaGu: thanks

packages/tc-ui/src/primitives/relationship/lib/relationship.jsx
kiyaGu: `ok, I think I follow. Do we have cypress tests for his scenario?`
Yes =>
https://github.com/Financial-Times/treecreeper/blob/6352deb1ef8352ca81cc55b447554fcf27cd3dc5/packages/tc-ui/src/pages/edit/__tests__/e2e-edit-record.cyp.js#L110

`Have you manually tested that the bad behaviour described in the comment doesn't happen after this change?`
I did and didn't see that behaviour happening ... it might be, in addition, because we have fixed the issue we had with Remove button responding to other clicks inside of `<label>` region

kiyaGu:
Thanks for reviewing thoroughly 👍

packages/tc-schema-sdk/server.js
```diff
@@ -1,7 +1,7 @@
 const { SDK } = require('./sdk');
 const readYaml = require('./lib/read-yaml');
 
-module.exports = new SDK({ init: false });
+module.exports = new SDK({ init: false, readYaml });
```
wheresrhys: I just had an idea. You can use `SDK.prototype.readYaml = readYaml;`, then you never have to pass it in explicitly here or in the tests as on any instance of the SDK `this.readYaml` will be inherited from the prototype

kiyaGu: giphy

PR #151, opened 2020-02-18
adds ability to lock a field globally

wheresrhys:
# Why
GDPR hub have a requirement that some fields on systems relating to GDPR compliance should not be editable via the Biz Ops UI. So we need a mechanism to lock some fields for all records of a given type

# What
- In the schema a field can have a `lockedBy` property, which lists all clients allowed to modify the field
- This is used to validate content submitted to the API
- This is also used to disable inputs in the UI
- Schema validation, api and UI e2e tests now exist for all 3

Also includes a little bit of housekeeping
- renames test-helpers/index.js to test-helpers/cypress.js
- removes the hard coding of the UIs client id as 'biz-ops-admin'

## Checklist
- [x] Schema change 
- [x] Schema documentation
- [x] Schema validation
- [x] API validation
- [x] API tests
- [x] Disable field in UI
- [x] Prevent submitting values from UI
- [x] Write cypress tests

### DIFF
--- a/packages/tc-ui/README.md
+++ b/packages/tc-ui/README.md
@@ -58,7 +58,7 @@ String to send to the API as the `Client-Id` header
 
 Either:
 
--   Function that is passed the object received as input, and should return an object containing additional headers to use when calling the treecreeper endpoints e.g. `api-key`, `client-user-id`.
+-   Function that is passed the object received as input, and should return an object containing additional headers to use when calling the treecreeper endpoints e.g. `Api-Key`, `Client-User-Id`.
 -   An object to be use directly as the headers
 
 ##### Subheader
--- a/packages/tc-ui/README.md
+++ b/packages/tc-ui/README.md
@@ -50,9 +50,15 @@ Base url for the treecreeper REST api
 
 Base url for the treecreeper GraphQL api
 
+##### clientId
+
+String to send to the API as the `Client-Id` header
+
 ##### apiHeaders
 
-Function that is passed the object received as input, and should return an object containing the headers to use when calling the treecreeper endpoints
+Either:
+- Function that is passed the object received as input, and should return an object containing additional headers to use when calling the treecreeper endpoints e.g. `api-key`, `client-user-id`.
+- An object to be use directly as the headers
 
 ##### Subheader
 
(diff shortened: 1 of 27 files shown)


packages/tc-ui/README.md
```diff
@@ -50,9 +50,16 @@ Base url for the treecreeper REST api
 
 Base url for the treecreeper GraphQL api
 
+##### clientId
+
+String to send to the API as the `Client-Id` header
```
kiyaGu: should this be `client-id`?

packages/tc-ui/README.md
```diff
@@ -50,9 +50,16 @@ Base url for the treecreeper REST api
...
+
 ##### apiHeaders
 
-Function that is passed the object received as input, and should return an object containing the headers to use when calling the treecreeper endpoints
+Either:
+
+-   Function that is passed the object received as input, and should return an object containing additional headers to use when calling the treecreeper endpoints e.g. `api-key`, `client-user-id`.
```
kiyaGu: is `client-user-id` supposed to be `client-id`? if not, what is the difference?

packages/tc-ui/src/pages/delete/__tests__/e2e-delete-record.cyp.js
```diff
@@ -10,7 +10,7 @@ const {
 	pickChild,
 	save,
 	resetDb,
-} = require('../../../test-helpers');
+} = require('../../../test-helpers/cypress');
```
kiyaGu: 👍

packages/tc-ui/README.md
wheresrhys: headers aren't case sensitive. Most documentation I've seen that mentions headers uses Capital-Case... don't know why, but that seems to be the accepted convention (e.g. check out any Fastly documentation)

packages/tc-ui/README.md
wheresrhys: Here's a documentation of what the value contains https://github.com/Financial-Times/treecreeper/blob/master/packages/tc-api-rest-handlers/README.md#metadata (although by the time it gets to this point in the app it is already converted from a header to a property of an object).

I will change the names above to Api-Key and Client-User-Id though as lowercasing them here is inconsistent with headers mentioned higher up

PR #152, opened 2020-02-19
record Annotations on relationships

kiyaGu:


packages/tc-api-rest-handlers/lib/relationships/write.js
```diff
@@ -49,22 +49,40 @@ WHERE related.code IN $${key}
 `;
 };
 
-const addPropsToQueries = (relationshipPropQueries, value) => {
-	// value: { code: 'node code', someString: 'some string'...}
-	Object.entries(value).forEach(([k, v]) => {
-		if (k !== 'code') {
-			// If no node matches the CASE expression, the expression returns a null.
-			// and no action will be taken
+const addPropsToQueries = (
```
wheresrhys: This looks like it's probably good :smile: 

but it does need tests in `api-rest-handlers/__tests__`

Need to check that
- a multiple choice property can be written to a relationship
- a boolean property can be written to a relationship
- different values can be written to 2 relationships at the same time
- (maybe) different values can be written to 2 relationships _of different types_ at the same time (though I'm not sure if the schema defines any relationships that will allow us to test this)

packages/tc-ui/src/primitives/relationship/lib/relationship-picker.jsx
```diff
@@ -128,6 +129,36 @@ class RelationshipPicker extends React.Component {
...
+	onChange(event) {
+		const { value, id, checked, dataset, type } = event.currentTarget;
+		const { parentCode } = dataset;
+		const propertyName = id.split('-')[1];
+		this.setState(prevState => {
+			const { selectedRelationships } = prevState;
+			const oldState = selectedRelationships.map(relationship => {
+				if (relationship.code === parentCode) {
+					// for multiple choice values
+					if (type === 'checkbox') {
```
wheresrhys: would it be possible to keep data manipulations specific to a component in that component? Because, say we need to add a new type of component that returns slightly different kind of data, then we won't need to edit relationship picker - we can just use the component. Biz Ops Admin already defines some custom components - there's a chance they won't work if relationship-picker doesn't understand the data they send, whereas if the pattern is to always massage the data first before forwarding it, then it shouldn't be a problem.  Something like
```jsx
// in the nested component
({onChange}) => {
const localOnChange = data => onChange(manipulate(data))
 return <input onChange={localOnChange} />
}
```

That's the theory anyway - it might not actually be a good idea - let's chat about it sometime

packages/tc-ui/src/primitives/relationship/lib/rich-relationship-properties.jsx
```diff
@@ -5,46 +5,39 @@ const {
...
+	const propertyfields = Object.entries(properties);
 
-	render() {
-		const { value, type, properties } = this.props;
-		const propertyfields = Object.entries(properties);
+	return propertyfields
+		.filter(
+			([, schema]) =>
+				// HACK: need to get rid of fields that are doing this
+				!schema.label.includes('deprecated') &&
```
wheresrhys: you can get rid of lines 15 & 16 - I have tidied up the deprecated stuff in the schema

PR #153, opened 2020-02-19
Fix relationship schema validator

wheresrhys:
# Why
The validation rules for the schema did not match up with what values the libraries expect to find for the case of definig direction of reflexive rich relationships

# What
- added an example reflexive rich relationship to the schema
- updated the schema validation tests to expect the right strings

### DIFF
--- a/packages/tc-schema-validator/tests/type-test-suite.js
+++ b/packages/tc-schema-validator/tests/type-test-suite.js
@@ -143,7 +143,7 @@ const propertyTestSuite = ({ typeName, properties, fieldsets }) => {
 					});
 					it('may have direction', () => {
 						if (config.direction) {
-							expect(config.direction).toMatch(/^from|to$/);
+							expect(config.direction).toMatch(/^incoming|outgoing$/);
 						}
 					});
 					it('can determine relationship direction explicitly', () => {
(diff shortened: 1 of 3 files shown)


packages/tc-schema-validator/tests/type-test-suite.js
```diff
@@ -420,50 +422,54 @@ const relationshipTestSuite = type => {
 			properties,
 		});
 
-		describe('from', () => {
-			it('has existing from type', () => {
+		describe('relationship endpoints', () => {
+			it('uses existing record type fro from', () => {
```
kiyaGu: typo `fro`

packages/tc-schema-validator/tests/type-test-suite.js
```diff
@@ -420,50 +422,54 @@ const relationshipTestSuite = type => {
...
-						prop => prop.direction === 'to',
-					);
-					expect(closeRel).toBeDefined();
-				}
-			});
-		});
 
-		describe('to', () => {
-			it('has existing to type', () => {
+			it('uses existing record type for to', () => {
```
kiyaGu: I found `'uses existing record type for to'` a bit vague ... are we saying type that already exists? how about `already defined`? 🤷‍♂

packages/tc-schema-validator/tests/type-test-suite.js
```diff
@@ -420,50 +422,54 @@ const relationshipTestSuite = type => {
...
-					prop => prop.type === type.name,
-				);
-				expect(propExists.length).toBeGreaterThan(0);
-				// If from and to type are the same, we should validate direction
-				if (from.type === to.type) {
-					const closeRel = propExists.find(
-						prop => prop.direction === 'from',
+
+			if (from.type !== to.type) {
+				it('from type makes use of this relationship type', () => {
```
kiyaGu: `from type makes use of this relationship type` when it read by itself makes sense but when you read it as `it from type makes use of this relationship type ` it doesn't

packages/tc-schema-validator/tests/type-test-suite.js
```diff
@@ -420,50 +422,54 @@ const relationshipTestSuite = type => {
...
+					const endType = types.find(t => t.name === to.type);
+					const propertiesUsingRelationshipType = Object.values(
+						endType.properties,
+					).filter(prop => prop.type === type.name);
+					expect(propertiesUsingRelationshipType.length).toBe(1);
+				});
+				it('to type makes use of this relationship type', () => {
+					const endType = types.find(t => t.name === from.type);
+					const propertiesUsingRelationshipType = Object.values(
+						endType.properties,
```
kiyaGu: didn't get this assertion - for example in `CuriousChild` I don't see where `prop.type === type.name` becomes true. am I missing something

PR #156, opened 2020-02-28
more sensible sdk initialisation

wheresrhys:
## Why?

We had a 'fix what annoys you' tech debt day. It annoys me that teh schema-sdk was fussy about whether you've called `init()` before you call `ready()`. This sometimes leads to hard to track down bugs and test failures.

## What?
One of the objects the SDK is composed of is a schema updater, and the `ready()` method is just a proxy through to `this.updater.ready()`. But previously the updater was only created on `init()`, so if `ready()` was called before `init()` you'd get an error. this was a pain because the ready method on the updater was clever enough to handle these sorts of race conditions, just the proxying wasn't.

This PR moves creating the updater into the SDK constructor, and then does a bit of juggling to ensure initialisation only happens when needed. I think it's more or less as sensible as before, but maybe now we're here the path to something more sensible will appear

### DIFF
--- a/packages/tc-api-express/index.js
+++ b/packages/tc-api-express/index.js
@@ -52,14 +52,13 @@ const getApp = async (options = {}) => {
 	router.use(bodyParsers);
 	router.use(restPath, restMiddlewares, getRestApi(options));
 	router.use(errorToErrors);
-
+	schema.init(schemaOptions);
 	const {
 		isSchemaUpdating,
 		graphqlHandler,
 		listenForSchemaChanges: updateGraphqlApiOnSchemaChange,
 	} = getGraphqlApi(options);
 
-	schema.init();
 	updateGraphqlApiOnSchemaChange();
 	updateConstraintsOnSchemaChange();
 
@@ -83,14 +82,12 @@ const getApp = async (options = {}) => {
 		availableEvents,
 	};
 
-	schema.init(schemaOptions);
 	updateGraphqlApiOnSchemaChange();
 	// avoids this running in every single spec file
 	// instead we explictly set up the constraints once before we start the test suite
 	if (!process.env.TREECREEPER_TEST) {
 		updateConstraintsOnSchemaChange();
 	}
-
 	await schema.ready();
 
 	return app;
--- a/packages/tc-schema-sdk/README.md
+++ b/packages/tc-schema-sdk/README.md
@@ -14,9 +14,11 @@ In many ways, this is the beating heart of Treecreeper. It consumes the schema f
 The package exports a singleton instance, and once initialised, `@financial-times/biz-ops-schema` can be required multiple times in the application. It should be initialised _once and once only per application_. It also exports a reference to the underlying `SDK` class, but this is only exposed for use by other packages' integration tests.
 
 ### Initialisation
+
 This is a little odd (and should be improved in future)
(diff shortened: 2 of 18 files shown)


packages/tc-api-express/index.js
```diff
@@ -52,14 +52,13 @@ const getApp = async (options = {}) => {
...
 	router.use(errorToErrors);
-
+	schema.init(schemaOptions);
 	const {
 		isSchemaUpdating,
 		graphqlHandler,
 		listenForSchemaChanges: updateGraphqlApiOnSchemaChange,
 	} = getGraphqlApi(options);
 
-	schema.init();
```
wheresrhys: this was superfluous

packages/tc-schema-validator/tests/enums.js
```diff
@@ -1,7 +1,9 @@
 /* global it, describe, expect */
 const { SDK } = require('@financial-times/tc-schema-sdk');
 
-const enums = new SDK().rawData.getEnums();
+const sdk = new SDK();
+sdk.init();
```
kiyaGu: What does the `init` function do if you don't pass it options? the tests will pass if we use `await sdk.ready();` instead as well...is it only for creating the updater?

wheresrhys: that is a good question. It calls `updater.configure()`, which reads local data or primes the updater to start fetching data, as appropriate. That behaviour is a bit weird. Maybe I should move polling initialisation into init too? Or... I'm not sure what's best.

I will document the behaviour a bit better for now, but don't want to make some API changes at the moment as I don't want to make your roll out of rich relationships harder

packages/tc-schema-sdk/README.md
```diff
@@ -14,6 +14,13 @@ In many ways, this is the beating heart of Treecreeper. It consumes the schema f
 The package exports a singleton instance, and once initialised, `@financial-times/biz-ops-schema` can be required multiple times in the application. It should be initialised _once and once only per application_. It also exports a reference to the underlying `SDK` class, but this is only exposed for use by other packages' integration tests.
 
 ### Initialisation
+This is a little odd (and should be improved in future)
+- When using local, static data (`schemaDirectory` or `schemaData` options below), the `init()` method populates the sdk with data immediately and its methods can be used to access the data immediately
+- When using remote data (`schemaBaseUrl`), no data is populated, and `schema.ready()` must be awaited before using the sdk's synchronous methods.
+
+Be aware of the idiosyncrasy above if oyu ever come across errors complaining that no data is available.
```
kiyaGu: typo `if you ever`

packages/tc-schema-sdk/README.md
```diff
@@ -15,6 +15,15 @@ The package exports a singleton instance, and once initialised, `@financial-time
 
 ### Initialisation
 
+This is a little odd (and should be improved in future)
+
+-   When using local, static data (`schemaDirectory` or `schemaData` options below), the `init()` method populates the sdk with data immediately and its methods can be used to access the data immediately
+-   When using remote data (`schemaBaseUrl`), no data is populated, and `schema.ready()` must be awaited before using the sdk's synchronous methods.
+
+Be aware of the idiosyncrasy above if oyu ever come across errors complaining that no data is available.
```
wheresrhys:
```suggestion
Be aware of the idiosyncrasy above if you ever come across errors complaining that no data is available.
```

PR #157, opened 2020-03-02
Upgrade to latest origami major versions

kiyaGu:
## Why?

-  Origami team have released a major version of all their [components](https://origami.ft.com/blog/2019/10/31/major-cascade/)

## What?

-  I have upgraded all the components to their latest version
- I have replaced `--o-colors-wasabi` with `--o-colors-jade` as wasabi is [removed](https://github.com/Financial-Times/o-colors/blob/master/MIGRATION.md#sass)

None


demo/cms/browser/main.css
```diff
@@ -136,7 +136,7 @@
 }
 
 .o-labels--green {
-	background: var(--o-colors-wasabi);
+	background: var(--o-colors-jade);
```
wheresrhys:
```suggestion
	background: var(--o-colors-jade);
	color: white;
```
Needed for contrast

demo/cms/components/subheader.jsx
```diff
@@ -16,7 +16,7 @@ const Subheader = ({ type, code, data }) => (
 					View runbook
 				</a>{' '}
 				<a
-					className="o-buttons o-layout__unstyled-element"
+					className="o-buttons o-buttons--secondary o-layout__unstyled-element"
```
wheresrhys: Is `o-layout__unstyled-element` still needed. I think it might be  athrowback to an older version of o-layout. @notlee Is it still a thing?

notlee: If `o-layout-typography` has not been applied to a parent element, `o-layout__unstyled-element` is not needed

kiyaGu: we use `o-layout-typography` in the parent element https://github.com/Financial-Times/treecreeper/blob/303450a27da90afc9ef913f40c1397e33cd7f9cd/packages/tc-ui/src/pages/view/template.jsx#L58

PR #163, opened 2020-03-04
remove final uses of biz ops from packages

wheresrhys:
## Why?
Final tidy up following import of code into treecreeper. 

## What?
Just a few find and replaces

### DIFF
--- a/packages/tc-markdown-parser/package.json
+++ b/packages/tc-markdown-parser/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@financial-times/tc-markdown-parser",
   "version": "0.0.0",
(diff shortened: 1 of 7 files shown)


packages/tc-markdown-parser/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@financial-times/tc-markdown-parser",
   "version": "0.0.0",
-  "description": "parse runbook.md defintitions which corresponds to biz-ops schema",
+  "description": "parse makrdown files and extract treecreeper records",
```
kiyaGu: just typo `markdown`

wheresrhys:
```suggestion
  "description": "parse markdown files and extract treecreeper records",
```

PR #164, opened 2020-03-05
fix bug detecting all error states for enum parser

wheresrhys:
## Why?

In runbook.md, tc-markdown-parser@0.4.0 was failing a test. The test tested if, when a parse error happened, the error message contained the correct line numbers. However in order to generate the problem node (which is the mechanism to respond with multiple parse errors) the error must be manually caught and converted. In the refactor when supporting multiple choice enums I'd inadvertently added a potential error state I wasn't catching.

## What?
Covers the case when the heading of a field which is supposed to contain an enum (or multi enum) has no content beneath it - checks for the length of child nodes and converts to a problem immediately if there are no children

Note - there are _no_ tests for error states in this package. Given that user experience depends on us covering them carefully I will add a ticket to create some

### DIFF
--- a/packages/tc-markdown-parser/lib/tree-mutators/coerce-treecreeper-properties-to-type.js
+++ b/packages/tc-markdown-parser/lib/tree-mutators/coerce-treecreeper-properties-to-type.js
@@ -78,7 +78,13 @@ const coerceEnumPropertyValue = (
 	{ propertyType: enumName, hasMany, enums },
 ) => {
 	let [subdocument] = node.children;
-
+	const validValues = Object.values(enums[enumName]);
+	if (!subdocument.children.length) {
+		return convertNodeToProblem({
+			node,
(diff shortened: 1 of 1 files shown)


packages/tc-markdown-parser/lib/tree-mutators/coerce-treecreeper-properties-to-type.js
```diff
@@ -78,7 +78,13 @@ const coerceEnumPropertyValue = (
 	{ propertyType: enumName, hasMany, enums },
 ) => {
 	let [subdocument] = node.children;
-
+	const validValues = Object.values(enums[enumName]);
+	if (!subdocument.children.length) {
```
kiyaGu: just curious, is it not possible to have a validation as well (or do we have one) when the schema is added to `biz-ops-schema` than when it is coerced? 🤔 😕

PR #167, opened 2020-03-12
Add icons

kiyaGu:
## Why?

We conducted a user testing for annotating relationships and this PR addresses some of the feedback we got

## What?

-   add tooltip and info icon for multiple choice options

### nested
Screenshot 2020-03-12 at 16 38 56

### not nested
Screenshot 2020-03-13 at 09 54 21

### DIFF
--- a/packages/tc-ui/src/primitives/multiple-choice/server.jsx
+++ b/packages/tc-ui/src/primitives/multiple-choice/server.jsx
@@ -14,10 +14,7 @@ const localOnChange = (event, currentValues, onChange) => {
 	onChange(propertyName, parentCode, values);
 };
 
-const getClass = isNested =>
-	isNested ? 'treecreeper-relationship-annotate__tooltip' : '';
-
-const getDescription = (checkboxValue, type, parentCode, isNested) => {
+const getDescription = (checkboxValue, type, parentCode) => {
 	const enumWithMeta = getEnums({ withMeta: true })[type];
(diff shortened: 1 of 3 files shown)


packages/tc-ui/src/primitives/multiple-choice/server.jsx
```diff
@@ -13,6 +14,42 @@ const localOnChange = (event, currentValues, onChange) => {
...
+			className="tooltip-container"
+			id={`tooltip-target-${parentCode}-${checkboxValue}`}
+		>
+			<span id={`tooltip-target-${parentCode}-${checkboxValue}`} />
+			<i
+				aria-label={`help for ${checkboxValue}`}
+				className="o-icons-icon o-icons-icon--info treecreeper-help-icon"
+			/>
+			<div
+				className={getClass(isNested)}
```
wheresrhys: I think we should use this when not nested too (we don't have any examples elsewhere in Biz Ops, but I think it's a nice feature)

wheresrhys: ...but I see in the css that implementing that would be extra work, so it's fine to leave it as it is for now

kiyaGu: @wheresrhys I have modified it to address your suggestion

wheresrhys: 🙌

PR #168, opened 2020-03-12
Number primitive fixes

wheresrhys:
## Why?

The UI didn't treat number values well at all
- It saves 0, but does not display 0 when in edit mode
- not sure it coerces the values sent by the form from strings into numbers
- The api doesn't throw an error if we try to save string values for int (it doesn't save the value to the database)
- need to also look into floats

## What?
Fixes all the problems above, and also adds a `someFloat` property to MainType so floats can be tested.

The problem with strings not throwing errors was quite interesting. The value was coerced using `Number(value)`, which converted strings to `NaN`, which the API interpreted as a null value. So now there is a check using `isNaN(value)` before attempting to coerce the string to a number


packages/tc-ui/src/primitives/number/__tests__/e2e-record-number.cyp.js
```diff
@@ -16,28 +17,100 @@ describe('End-to-end - record Number type', () => {
...
+			cy.get('input[name=someFloat]').type(someInteger);
+			save();
+
+			cy.get('#code').should('have.text', code);
+			cy.get('#someString').should('have.text', someString);
+			cy.get('#someFloat').should('have.text', String(someInteger));
+		});
+
+		it('accepts floats', () => {
+			cy.get('input[name=someFloat]').type('0.5');
```
kiyaGu: same here `type(0.5)`

packages/tc-ui/src/primitives/number/__tests__/e2e-record-number.cyp.js
```diff
@@ -16,28 +17,100 @@ describe('End-to-end - record Number type', () => {
...
+				'Oops. Could not create MainType record for e2e-demo',
+			);
+			cy.get('.o-message__content-additional').should(
+				'contain',
+				`Invalid value \`haha\` for property \`someInteger\` on type \`MainType\`: Must be a finite integer`,
+			);
+		});
+
+		it('saves and redisplays zero', () => {
+			cy.get('input[name=someInteger]').type('0');
```
kiyaGu: same here `type(0)`

packages/tc-ui/src/primitives/number/__tests__/e2e-record-number.cyp.js
```diff
@@ -16,28 +17,100 @@ describe('End-to-end - record Number type', () => {
...
-				7}\` for property \`someInteger\` on type \`MainType\`: Must be a finite integer`,
-		);
-	});
+			cy.get('#code').should('have.text', code);
+			cy.get('#someString').should('have.text', someString);
+			cy.get('#someInteger').should('have.text', String(someInteger));
+		});
+
+		it('rejects floats', () => {
+			cy.get('input[name=someInteger]').type('0.5');
```
kiyaGu: why not `type(0.5)` as we are testing for float

wheresrhys: I was guessing that cypress can only type strings. When the form is submitted they are sent as strings anyway

PR #173, opened 2020-03-23
Interesting related codes

wheresrhys:
## Why?

We'd never tried writing rich relationship props onto relationships to things with codes using symbols before. It didn't work. https://trello.com/c/DrJnmS07/313-cannot-write-rich-relationships-when-code-contains-special-characters

image

## What?

- Defines a new type that allows any string as a code in order to facilitate testing
- Takes a more aggressive approach to sanitising codes - only allows letters, digits and underscore
- Uses the index in the variable name in order to ensure uniqueness

### DIFF
--- a/example-schema/relationships/OddThing.yaml
+++ b/example-schema/relationships/OddThing.yaml
@@ -1,0 +1,13 @@
+name: OddThing
+from:
+  type: MainType
+  hasMany: false
+to:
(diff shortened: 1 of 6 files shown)


wheresrhys:
Gonna merge now as need to not be distracted by bugs tomorrow

PR #175, opened 2020-03-24
for runbooks: set link to the dependency/dependent System point to its runbook not biz-ops page

kiyaGu:
## Why?

-   In runbooks we want the link to the dependency/dependent to point to that system's runbook not biz-ops page

## What?

-  Passed a flag that can identify whether to point to the dependency/dependent runbook or biz-ops page when rendering the link to the dependency/dependent system.

My initial implementation was done by sending a href like `https://runbooks.in.ft.com/${encodeURIComponent(code)}` this can work for a single relationship but if there are more than one relationships this will not work as when we render the relationships, using the ViewComponent, we send an array of dependencies/dependents each with its own code.

### DIFF
--- a/packages/tc-ui/src/lib/components/structure.jsx
+++ b/packages/tc-ui/src/lib/components/structure.jsx
@@ -139,11 +139,21 @@ const MetaProperties = ({ data, isCreate }) => {
 	);
 };
 
-const LinkToRecord = ({ id, type, value: { name, code } }) => (
-	<a id={id} href={`/${type}/${encodeURIComponent(code)}`}>
-		{name || code}
-	</a>
-);
+const LinkToRecord = ({
+	id,
+	type,
+	value: { name, code },
+	isRunbooksSystemLink,
+}) => {
+	const href = isRunbooksSystemLink
+		? `https://runbooks.in.ft.com/${encodeURIComponent(code)}`
+		: `/${type}/${encodeURIComponent(code)}`;
+	return (
+		<a id={id} href={href}>
+			{name || code}
+		</a>
+	);
+};
 
 module.exports = {
 	LinkToRecord,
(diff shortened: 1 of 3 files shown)


packages/tc-ui/src/lib/components/structure.jsx
```diff
@@ -139,11 +139,21 @@ const MetaProperties = ({ data, isCreate }) => {
...
-		{name || code}
-	</a>
-);
+const LinkToRecord = ({
+	id,
+	type,
+	value: { name, code },
+	isRunbooksSystemLink,
+}) => {
+	const href = isRunbooksSystemLink
```
wheresrhys: It's important to remember that treecreeper shouldn't know anything about biz ops, runbooks etc, so referring to runbooks in the codebase is a no no. Think more in terms of naming things something like 'alternativeLinkFormat`. Once you restrict yourself to not allowing any mention of runbooks, or specifying the actual link format in tc-ui code, what approaches are left that you can take?

kiyaGu: I have replaced it with `alternativeHostname` and we send this from runbooks as `https://runbooks.in.ft.com` if type is System instead of a boolean value

packages/tc-ui/src/lib/components/structure.jsx
```diff
@@ -139,17 +139,10 @@ const MetaProperties = ({ data, isCreate }) => {
...
-	id,
-	type,
-	value: { name, code },
-	alternativeHostname,
-}) => {
-	const href = alternativeHostname
-		? `${alternativeHostname}/${encodeURIComponent(code)}`
-		: `/${type}/${encodeURIComponent(code)}`;
+const LinkToRecord = ({ id, type, value: { name, code }, linkGenerator }) => {
+	const href = linkGenerator({ type, code });
```
wheresrhys:
```suggestion
	const href = linkGenerator({ type, ...value });
```
I think this is better as it gives more flexibility. It's also compatible with what you've already done.

PR #177, opened 2020-03-25
split MVR out of main type

kiyaGu:
## Why?

https://trello.com/c/wNPVcp9b/306-split-mvr-out-of-main-type

## What?

-  created MVR type
- create more test helper functions that can be used for creating types
- refactored test to make e2e test faster

### Before
Screenshot 2020-03-27 at 17 15 40

### After

Screenshot 2020-03-27 at 17 40 59


packages/tc-ui/src/pages/delete/__tests__/e2e-delete-record.cyp.js
```diff
@@ -15,26 +15,26 @@ const {
 describe('End-to-end - delete record', () => {
 	beforeEach(() => {
 		cy.wrap(resetDb()).then(() => {
-			cy.visit(`/MainType/create`, {
+			cy.visit(`/MVRType/create`, {
```
wheresrhys: this can still be maintype, no? What's the reason for using MVR?

packages/tc-ui/src/pages/edit/__tests__/e2e-edit-MVR-record.cyp.js
```diff
@@ -0,0 +1,77 @@
+const { code, someString } = require('../../../test-helpers/mainTypeData.json');
```
wheresrhys: I don't think we need this as well as the edit record tests for MainType. Are you testing anything different here compared to the MainType one?

packages/tc-ui/src/pages/view/__tests__/e2e-view-complete-record.cyp.js
```diff
@@ -1,135 +1,49 @@
...
 
-	it('can create MainType record', () => {
-		populateMinimumViableFields(code);
-		save();
-
-		cy.url().should('contain', `/MainType/${code}`);
-		cy.get('#record-title a span:first-child').should(
-			'have.text',
-			'MainType: e2e-demo',
+	it('can display MVRType record', () => {
```
wheresrhys: don't need this if we have the MainType test too

packages/tc-ui/src/pages/edit/__tests__/e2e-edit-MVR-record.cyp.js
kiyaGu: it was only `it('can not edit code of record', () => { ...}` will move that test to MainType

kiyaGu:
> I think there's a handful of tests you can remove, but otherwise looks good.
> 
> 2m19s saved is good, but I'm surprised it hasn't shaved even more off the execution time.
> 
> I wonder... would it be possible to split the cypress task into 2 jobs - cypress-page and cypress-primitives, and make sure we use different values for `code` in both. Then we could run both containers in parallel

@wheresrhys not sure which approach I should take between the following 
 1. after removing some of the test you suggested in your review 

Screenshot 2020-03-30 at 09 55 51

2. after splitting the tests (pages and primitives) including the changes in #1

Screenshot 2020-03-30 at 11 07 39

the difference is ~ 29 sec

PR #178, opened 2020-03-25
fix demo app sub-header

kiyaGu:
## Why?

- The current [implementation](https://github.com/Financial-Times/treecreeper/blob/303450a27da90afc9ef913f40c1397e33cd7f9cd/demo/cms/components/subheader.jsx#L3) of demo app's `Subheader` component is not working properly when we hookup treecreeper with Biz Ops. The reason for this is when we distructure properties form the prop `type` is undefined as it is not defined in the root but as
```
{
    schema: {
        type: "System"
        ...
    }
}
```

## What?

-   replaced the `Subheader` component with the one that is implemented in Biz-ops-admin

### Before
Screenshot 2020-03-25 at 12 48 36

### After
Screenshot 2020-03-25 at 12 50 18

### DIFF
--- a/demo/cms/components/subheader.jsx
+++ b/demo/cms/components/subheader.jsx
@@ -1,51 +1,111 @@
 const React = require('react');
 
-const Subheader = ({ type, code, data }) => (
+const newTabLinkProps = url => {
+	if (String(url).startsWith('http')) {
+		return { rel: 'noopener noreferrer', target: '_blank' };
+	}
+	return {};
+};
+
+const AlternateViewLink = ({ url, ariaLabel, text }) => (
+	<>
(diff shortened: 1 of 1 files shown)


demo/cms/components/subheader.jsx
```diff
@@ -1,51 +1,111 @@
...
+		>
+			{text}
+		</a>{' '}
+	</>
+);
+
+const Subheader = ({
+	schema: { type },
+	data: { code, url, _lockedFields, runbookMdUrl, lastReleaseTimestamp },
+	data,
```
wheresrhys: I have never seen this pattern before - I learned something new today

PR #180, opened 2020-03-30
allow passing extra resolvers and schema extensions

kiyaGu:
## Why?

-   Allow passing in extra resolvers and schema extensions as options in to [tc-api-graphql](https://trello.com/c/sWJrp982/282-allow-passing-in-extra-resolvers-and-schema-extensions-as-options-in-to-tc-api-graphql) 

## What?

-   added `options` argument 
```
{
 ...,
 options: {
  schema: [...],
  resolvers: {...}
}
```
- [ ] update documentation

### DIFF
--- a/packages/tc-api-graphql/README.md
+++ b/packages/tc-api-graphql/README.md
@@ -48,7 +48,7 @@ An [optional] object value for adding extra resolvers and typeDefs. It should ha
         someEnum: AnEnum
       }`,
       `extend type MainType {
-        extended: ExtendedType @neo4j_ignore
+        extended: ExtendedType @neo4j_ignore *
         }`,],
     resolvers: {
       MainType: {
@@ -58,6 +58,7 @@ An [optional] object value for adding extra resolvers and typeDefs. It should ha
   }
 }
 ```
+*\* **@neo4j_ignore** must be added so that it is not used during the schema augmentation process*
 
 ## Example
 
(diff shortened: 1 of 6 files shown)


packages/tc-api-graphql/lib/get-augmented-schema.js
```diff
@@ -34,8 +34,18 @@ const getDocumentResolvers = () => {
 	return typeResolvers;
 };
 
-const getAugmentedSchema = ({ documentStore }) => {
+const getAugmentedSchema = ({ documentStore, options }) => {
+	const resolvers = documentStore ? getDocumentResolvers() : {};
 	const typeDefs = getGraphqlDefs();
+
+	if (options.schema) {
+		typeDefs.push(options.schema);
```
wheresrhys:
```suggestion
		typeDefs.push(...options.typeDefs);
```
I think you need to spread if it's an array?

Also prefer typeDefs to schema as typeDefs is what the graphql libraries use

packages/tc-api-graphql/README.md
```diff
@@ -58,6 +58,7 @@ An [optional] object value for adding extra resolvers and typeDefs. It should ha
   }
 }
 ```
+*\* **@neo4j_ignore** must be added so that it is not used during the schema augmentation process*
```
wheresrhys: :+1:

packages/tc-api-graphql/index.js
```diff
@@ -7,13 +7,14 @@ const getGraphqlApi = ({
 	documentStore,
 	republishSchema,
 	republishSchemaPrefix = 'api',
+	options = {},
```
wheresrhys: I didn't notice something earlier. I think it makes more sense to pass them in as top level properties on the options. I can't think of any good reason why they should be wrapped in another object. Can you?

```suggestion
	typeDefs = [],
	resolvers = {},
```

PR #182, opened 2020-03-31
add command that can hookup treecreeper with biz-ops

kiyaGu:
## Why?

-   we need to have an option to hookup treecreeper with biz-ops 

## What?

-   added `run-biz-ops` that can hookup treecreeper with treecreeper

https://github.com/Financial-Times/treecreeper/blob/7219aeb8916d25a671f4c931f03ec76d52b813e0/packages/tc-schema-sdk/lib/updater.js#L25
I made it to default to `stall` if updateMode is not passed or TREECREEPER_SCHEMA_URL is not set What do you think?


Makefile
```diff
@@ -91,15 +91,16 @@ run:
...
 
 cypress-page:
 	start-server-and-test "make run-app" http-get://localhost:8888/MainType/create "make cypress-run-page"
 
 cypress-primitives:
 	start-server-and-test "make run-app" http-get://localhost:8888/MainType/create "make cypress-run-primitives"
 
-
+run-biz-ops: build-statics
+	NEO4J_BOLT_URL=${BIZ_OPS_BOLT_URL} TREECREEPER_SCHEMA_URL=${SCHEMA_BASE_URL} SCHEMA_UPDATE_Mode=poll PROJECT_NAME=biz-ops-api PRODUCT_NAME=biz-ops nodemon --inspect demo/api.js
```
wheresrhys:
```suggestion
	NEO4J_BOLT_URL=${BIZ_OPS_BOLT_URL} TREECREEPER_SCHEMA_URL=${SCHEMA_BASE_URL} PROJECT_NAME=biz-ops-api PRODUCT_NAME=biz-ops nodemon --inspect demo/api.js
```

I guess you don't ned to pass in update mode as it defaults in the library now?

packages/tc-schema-sdk/lib/updater.js
```diff
@@ -22,7 +22,8 @@ class SchemaUpdater {
 		schemaDirectory = process.env.TREECREEPER_SCHEMA_DIRECTORY,
 		schemaData,
 	} = {}) {
-		this.updateMode = updateMode;
+		this.updateMode =
+			updateMode || process.env.TREECREEPER_SCHEMA_URL ? 'poll' : 'stale';
```
wheresrhys:
```suggestion
			updateMode || (schemaBaseUrl ? 'poll' : undefined');
```

Not sure if this will pass linting now, but is this more the behaviour we need? Because
- if the user has passed in a schemaBaseUrl manually, we probably still want to have the same behaviour as if they used an env var (at least, that feels more predictable to me)
- if there is a directory/data instead of url, then we don't need any update mode

PR #183, opened 2020-03-31
Ensure string validations match on full strings

katsuraku:
## Why?

-   The string pattern definitions in the schema should all be for entire strings, but previously there were some rules in Biz ops that lacked this. So we need to add a test in the schema-validator test suite
https://trello.com/c/xLpMuomO/324-foce-string-patterns-to-begin-and-end

-   What's the main motivation of this work? New feature, bug fix, tech debt ... ?
Listed as a bug on the Trello ticket

## What?

-   Added a new test that checks string validation regexes to make sure they always have a ^ at the beginning and a $ at the end
-  Updated the COST_CENTRE regex pattern, so that it has the missing $ at the end of it
-  Witnessed the new test failing when removing the ^ or $ from one of the patterns.

I'm in the midst of figuring out why I've got lots of failing tests locally. I've witnessed the expected failure of the new test as described above, and the tests pass in circle. So I think this is safe.


lucas42:
If the regexes are always going to start and end in `^` and `$`, then why doesn't treecreeper just add those in automatically and have the schema define the bit in the middle?

wheresrhys:
> If the regexes are always going to start and end in ^ and $, then why doesn't treecreeper just add those in automatically and have the schema define the bit in the middle?

I did think about that, but felt the less magic, more explicit declaration of the full regex would be clearer

example-schema/string-patterns.yaml
```diff
@@ -2,7 +2,7 @@ NODE_TYPE: '^(?=.{2,64}$)([A-Z][a-z]+)+$' # NodeType
 CODE: '^(?=.{2,64}$)[a-z0-9]+(?:-[a-z0-9]+)*$' # system-code.1
 RELATIONSHIP_NAME: '^(?=.{2,64}$)[A-Z][A-Z_]*[A-Z]$' # HAS_A_RELATIONSHIP
 ATTRIBUTE_NAME: '^(?=.{2,64}$)[a-z][a-zA-Z\d]+$' # attributeName
-COST_CENTRE: '^(?=.{2,64}$)[A-Z][A-Z]\d\d\d' # XT100
+COST_CENTRE: '^(?=.{2,64}$)[A-Z][A-Z]\d\d\d$' # XT100
```
wheresrhys: most of these patterns can actually be removed from this repo - an artefact from when the schema was copied over from Biz Ops Schema. 

I'll comment on each line (including this one) that can be removed, but there might be more
```suggestion
```

wheresrhys: actually, I can't comment on anythnig hat's more than 5 lines away from your changes, but repo name, package name and the max length ones can definitely go. Probably everything apart from the top 4 lines, so try searcing in the `example-schema` folder to see if any of the others are mentioned. If not, you can delete them

katsuraku:
Unused string patterns removed

katsuraku:
just doing some messing about to find out why the verify stage of the ci build is failing...

katsuraku:
@wheresrhys @lucas42 i forgot about this - is it ok to merge?

PR #221, opened 2020-04-22
Decom component

katsuraku:
@wheresrhys The test is not very good but it might be all I can add.
The components that we have in the cypress tests are all determined by MainType.yaml, so I don't think I should be adding another custom thing in there to force it to render.
The logic that decides whether the decom button shows up is not just in the decom component, but also in the demo app in primitives.jsx, which passes the custom component in.
So all that renders on the cypress browser is the stuff from MainType.yaml, and unconditionally rendered divs, like the 'additionale-edit-hydration-container' (but not other component types).

This is what the stuff rendering looks like 
Screenshot 2020-04-22 at 11 47 55


demo/cms/components/decomButton.jsx
```diff
@@ -0,0 +1,10 @@
+const React = require('react');
+
+const DecomButton = props => (
+	<div className="decommission-override">
+		Stand-in decom component for example app
+		{props.entireRecord.lifecycleStage}
```
wheresrhys, line 6:
```suggestion
		{props.entireRecord.someString}
```

As you'll be working with the example schema even though you can't access lifecycleStage, you can use other properties defined on MainType to test that the values are getting through

katsuraku: Ah what I meant was, I can't even access .decommission-override. 
I've just pushed up a failing test to show what I mean. 
`cy.get('.decommission-override').should('exist'); // fails`

katsuraku: That's what I meant about the .yaml file. I think that dictates what renders on the cypress version, but divs which aren't inside a named component also render out, which is how I get access to .additional-edit-hydration-container
I can test that that div exists but that's it, unless there's a way I can't see to get other things to render.

demo/cms/components/primitives.jsx
```diff
@@ -95,6 +96,11 @@ const RelationshipAnnotator = ({ type, value }) => (
 primitives.Relationship.setRelationshipAnnotator(RelationshipAnnotator);
 
 module.exports = {
+	SystemLifecycle: {
```
wheresrhys, line 99:
```suggestion
	AnEnum: {
```
The keys of this object need to be the names of actual primitive types or enums.

**Warning!** may end up breaking some existing tests so might be more straightforward to define a new type of enum in enums.yaml and add another property to MainType.yaml that uses this new enum

PR #233, opened 2020-04-29
Enable markdown in the field descriptions!

GeoffThorpeFT:
## Why?

We appear to have `markdown` syntax in our biz ops schema field descriptions but we just display them as raw text.

## What?

Call `showdown` to convert field descriptions from markdown to html.

Untested - as `make install` for this monorepo isnt completing on my windows machine at the moment

## Screenshots / Images

Add any Screenshots or Images here to show any change/s

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -479,6 +479,17 @@ ignore:
     - '@financial-times/tc-ui > showdown > yargs > yargs-parser':
         reason: no fix yet
         expires: '2020-04-23T12:54:36.605Z'
-
+  SNYK-JS-LODASH-567746:
+    - next-metrics > lodash:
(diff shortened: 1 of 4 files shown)


wheresrhys:
Yes to the functionality, but would be good to put the code that already implements thsi conversion elsewhere into a reusable module and use that instead https://github.com/Financial-Times/treecreeper/blob/master/packages/tc-ui/src/primitives/large-text/server.jsx#L12

GeoffThorpeFT:
@wheresrhys  Thanks for the suggestion; hadn't seen that other bit of code.  I've applied a quick refactor.

PR #235, opened 2020-04-30
Tidied makefile

wheresrhys:
## Why?

Tasks were poorly organised. Also webpack had stopped running in watch mode locally because of some changes made to support cypress

## What?
- Reorders and groups into Install, Build & Test sections
- Documents what the various cypress tasks are used for
- Defines a new `run-app-biz-ops` task so we can use `concurrently` to run that task in parallel with webpack in watch mode
- _Hopefully_ doesn't break anything

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,7 +104,7 @@ test:
 		then NEO4J_BOLT_URL=${LOCAL_BOLT_URL} TREECREEPER_TEST=true TREECREEPER_SCHEMA_DIRECTORY=example-schema DEBUG=true TIMEOUT=500000 \
 			jest --config="./jest.config.js" "${pkg}.*__tests__.*${spec}.*.spec.js" --testEnvironment=node --watch; \
 		else NEO4J_BOLT_URL=${LOCAL_BOLT_URL} TREECREEPER_TEST=true TREECREEPER_SCHEMA_DIRECTORY=example-schema \
-			jest --config="./jest.config.js" "__tests__.*/*.spec.js" --testEnvironment=node --maxWorkers=1 --ci --reporters=default --reporters=jest-junit --detectOpenHandles --forceExit; \
+			jest --config="./jest.config.js" "__tests__.*/*.spec.js" --testEnvironment=node --runInBand --ci --reporters=default --reporters=jest-junit --detectOpenHandles --forceExit; \
 	fi
 
 ## Cypress stuff used in CI
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,8 @@ verify:
 unprepublish:
 	sed s/"dist\/"/"src\/"/ packages/tc-ui/package.json > tmp && mv tmp packages/tc-ui/package.json
 
+## Note - no need to clean up ./packages/**/package-lock.json as they are installed lockless
+## https://github.com/Financial-Times/treecreeper/blob/master/package.json#L59
 clean-deps: unprepublish
 	rm -rf packages/*/node_modules
 	rm -rf node_modules
--- a/Makefile
+++ b/Makefile
@@ -20,11 +20,14 @@ node_modules/@financial-times/rel-engage/index.mk:
 LOCAL_BOLT_URL=bolt://localhost:7687
 BIZ_OPS_BOLT_URL=${NEO4J_BOLT_URL}
 NEO4J_VERSION=3.5.0
-
-
 PRODUCT_NAME=biz-ops
 PROJECT_NAME=biz-ops-api
 
+
+.PHONY: test
+
+
+# Install and dependency management
 env:
 	echo "No secret environment variables needed in test"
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -33,79 +36,56 @@ verify:
 unprepublish:
 	sed s/"dist\/"/"src\/"/ packages/tc-ui/package.json > tmp && mv tmp packages/tc-ui/package.json
 
+clean-deps: unprepublish
+	rm -rf packages/*/node_modules
```
i-like-robots, line 40: Does this need to clean up lock files in the packages too?

Makefile
```diff
@@ -33,79 +36,56 @@ verify:
...
-
-run-app: build-statics
-	NEO4J_BOLT_URL=${LOCAL_BOLT_URL} TREECREEPER_TEST=true TREECREEPER_SCHEMA_DIRECTORY=example-schema nodemon --inspect demo/api.js
+# Building and running
 
+## Builds browser files for tc-ui (in watch mode when local)
 build-statics:
 	@if [ -z $(CI) ]; \
-		then $(info Webpack bundling modules ...) ./node_modules/.bin/webpack --mode=development; \
+		then $(info Webpack bundling modules ...) ./node_modules/.bin/webpack --mode=development --watch; \
```
i-like-robots, line 55: `npx webpack`?

Makefile
```diff
@@ -117,8 +97,52 @@ run-test-db:
...
-	rm -rf packages/*/node_modules
-	rm -rf node_modules
-	rm package-lock.json
-	npm install
+test:
+	@if [ -z $(CI) ]; \
+		then NEO4J_BOLT_URL=${LOCAL_BOLT_URL} TREECREEPER_TEST=true TREECREEPER_SCHEMA_DIRECTORY=example-schema DEBUG=true TIMEOUT=500000 \
+			jest --config="./jest.config.js" "${pkg}.*__tests__.*${spec}.*.spec.js" --testEnvironment=node --watch; \
+		else NEO4J_BOLT_URL=${LOCAL_BOLT_URL} TREECREEPER_TEST=true TREECREEPER_SCHEMA_DIRECTORY=example-schema \
+			jest --config="./jest.config.js" "__tests__.*/*.spec.js" --testEnvironment=node --maxWorkers=1 --ci --reporters=default --reporters=jest-junit --detectOpenHandles --forceExit; \
```
i-like-robots, line 105: I wonder if `--runInBand` is faster than `maxWorkers=1` as this would avoid creating a worker pool?

Makefile
```diff
@@ -117,8 +97,52 @@ run-test-db:
...
+### Runs tests for primitive components (assumes the app is running in another terminal)
+cypress-run-primitives:
+	TREECREEPER_TEST=true TREECREEPER_SCHEMA_DIRECTORY=example-schema \
+	cypress run --spec "packages/tc-ui/src/primitives/**/__tests__/**.cyp.js"
+
+# Deploy
+
+prepublish:
+	babel packages/tc-ui/src -D --out-dir packages/tc-ui/dist
+	sed s/"src\/"/"dist\/"/ packages/tc-ui/package.json > tmp && mv tmp packages/tc-ui/package.json
```
i-like-robots, line 144: Should this be a `prepare` script within the `tc-ui` package itself?

Makefile
wheresrhys: nope - the install step for packages skips package-lock https://github.com/Financial-Times/treecreeper/blob/master/package.json#L59 (following your advice from a few months ago)

Makefile
```diff
@@ -33,79 +36,56 @@ verify:
 unprepublish:
 	sed s/"dist\/"/"src\/"/ packages/tc-ui/package.json > tmp && mv tmp packages/tc-ui/package.json
 
+clean-deps: unprepublish
```
wheresrhys, line 39:
```suggestion
## Note - no need to clean up ./packages/**/package-lock.json as they are installed lockless
## https://github.com/Financial-Times/treecreeper/blob/master/package.json#L59
clean-deps: unprepublish
```

Makefile
wheresrhys: just webpack will do as node_modules should be in path
```suggestion
		then $(info Webpack bundling modules ...) webpack --mode=development --watch; \
```

Makefile
wheresrhys: Happy to give it a go
```suggestion
			jest --config="./jest.config.js" "__tests__.*/*.spec.js" --testEnvironment=node --runInBand --ci --reporters=default --reporters=jest-junit --detectOpenHandles --forceExit; \
```

Makefile
wheresrhys: Probably, but there might be a good reason to keep separate... something something... local dev... mumble mumble

Will leave as is for now

i-like-robots: I did see that this tweaks the entry point of the package manifest - I assume so local linking can consume from source instead of rebuilding each change? One for a separate PR anyway.

Issue #206, opened 2020-04-20
Index all canIdentify fields


wheresrhys:
When running `CALL db.indexes();` in neo4j only 23 are retrieved, but there should be [many more](https://github.com/Financial-Times/biz-ops-schema/search?q=canIdentify&unscoped_q=canIdentify), as stipulated by the use of `canIdentify` on properties in the schema.

In tc-api-db-manager [only properties marked with `unique` are indexed](https://github.com/Financial-Times/treecreeper/blob/master/packages/tc-api-db-manager/index.js#L38)... this should most likely be updated to go by `canIdentify` (I can't remember, but maybe an earlier version of the schema used unique instead of canIdentify)

wheresrhys:
The cypher command to create an index is e.g. `CREATE INDEX ON :Person(email)`

wheresrhys:
Just remembered I was gonna tell you about the release process
1. Merge to master
2. Create a semver tag - this patches a bug so it'll be a patch release (bring it up in standup if you're not familiar with semver)
3. Wait for the circleci build to complete the publish step
4. Clone the biz-ops-api repo and create a branch
5. Update it's package.json to require _all_ `tc-*` packages at the new version
6. run `make install`
7. Because the app has not been deployed for a while, I recommend running snyk wizard too because theres' bound to be some vulnerability that CI will catch
8. Once the PR's passed CI and approved, merge it
9. When circleci has finished head over to https://dashboard.heroku.com/pipelines/d5deb97d-5fa2-45f2-99fa-cd155328320d (hopefully you have access). Wait for the staging build to finish
10. Go to https://dashboard.heroku.com/apps/biz-ops-api-staging  Click on the grapheneDB button and make your way to the neo4j browser (you will probably need to go back to the env vars bit of the heroku dashboard to copy username and password for the db)
11. Once you're in you can run `call db.indexes()` to make sure everything exists as expected
12. Go to https://biz-ops-test.in.ft.com/api-explorer and make sure the graphql api works ok
13. Return to the pipeline dashboard and click deploy to production

PR #239, opened 2020-05-07
Create index can identify

Leigan0:
## Why?

To resolve [issue](https://github.com/Financial-Times/treecreeper/issues/206)

#### Problem 
When running CALL db.indexes(); in neo4j only 23 are retrieved, but there should be many more, as stipulated by the use of canIdentify on properties in the schema.

In tc-api-db-manager only properties marked with unique are indexed... this should most likely be updated to go by canIdentify 

-   What's the main motivation of this work? bug fix

## What?

I added a condition for to check for canIdentify property. If this is present when unique is false the code marks this field for index creation. I have also updated the code to retrieve current indexes, if the index already exists nothing further will happen, if the index is not present a new index will be created for the field with canIdentify. 

When implementing I noticed that the tests checking that existing constraints were not created again did not actually check that behaviour so I have updated these also. 

I have also added on check for amount of calls expected to `db.run` whilst this is not all that clear on first glance, it should help highlight incorrect passes when uses the .not. expectation, as if there is a typo / error in `.not` expectation the test would still pass.  

_Would appreciate a second pair of eyes on the test_  to confirm this makes sense with intended behaviour.

### DIFF
--- a/packages/tc-api-db-manager/__tests__/db-manager.spec.js
+++ b/packages/tc-api-db-manager/__tests__/db-manager.spec.js
@@ -60,6 +60,7 @@ describe('creating db constraints and indexes', () => {
 		expect(dbRun).toHaveBeenCalledWith(
 			'CREATE CONSTRAINT ON (s:Dog) ASSERT s.tail IS UNIQUE',
 		);
+		expect(dbRun).toHaveBeenCalledTimes(5);
 	});
 	// ignore this test until we upgrade to enterprise adiition - community edition doesn't have exists()
 	it.skip("creates an existence constraint if it doesn't exist", async () => {
@@ -86,8 +87,9 @@ describe('creating db constraints and indexes', () => {
 		]);
 		await initConstraints();
 		expect(dbRun).not.toHaveBeenCalledWith(
-			'CONSTRAINT ON (s:Dog) ASSERT s.nose IS UNIQUE',
+			'CREATE CONSTRAINT ON (s:Dog) ASSERT s.nose IS UNIQUE',
 		);
+		expect(dbRun).toHaveBeenCalledTimes(4);
 	});
 
 	// ignore this test until we upgrade to enterprise adiition - community edition doesn't have exists()
@@ -101,8 +103,9 @@ describe('creating db constraints and indexes', () => {
 		]);
 		await initConstraints();
 		expect(dbRun).not.toHaveBeenCalledWith(
-			'CONSTRAINT ON (s:Dog) ASSERT exists(s.nose)',
+			'CREATE CONSTRAINT ON (s:Dog) ASSERT exists(s.nose)',
 		);
+		expect(dbRun).toHaveBeenCalledTimes(4);
 	});
 
 	it('creates a index for canIdentify property if it doesnt exist', async () => {
@@ -115,6 +118,7 @@ describe('creating db constraints and indexes', () => {
 		]);
 		await initConstraints();
 		expect(dbRun).toHaveBeenCalledWith('CREATE INDEX ON :Dog(tail)');
+		expect(dbRun).toHaveBeenCalledTimes(5);
 	});
 
 	it('doesnt create an index for for canIdentify property if it does exist', async () => {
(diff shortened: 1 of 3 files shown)


packages/tc-api-db-manager/index.js
```diff
@@ -27,20 +28,39 @@ const initConstraints = async () => {
...
-								];
+									// 	// required &&
+									// 	// 	`CONSTRAINT ON (s:${typeName}) ASSERT exists(s.${propName})`
+									// ];
+								}
+								if (canIdentify) {
+									return [
+										`INDEX ON :${typeName}(${propName})`,
+									];
+								}
```
wheresrhys, line 63: does it need to `return []` by default in order to avoid `desiredConstraintsAndIndexes` containing `undefined` items?

packages/tc-api-db-manager/__tests__/db-manager.spec.js
```diff
@@ -60,37 +78,91 @@ describe('creating db constraints', () => {
...
+		});
+		schema.getTypes.mockReturnValue([
+			{ name: 'Dog', properties: { tail: { canIdentify: true } } },
+		]);
+		await initConstraints();
+		expect(dbRun).toHaveBeenCalledWith('CREATE INDEX ON :Dog(tail)');
+		expect(dbRun).toHaveBeenCalledTimes(5);
+	});
+
+	it('doesnt create an index for for canIdentify property if it does exist', async () => {
```
wheresrhys, line 124:
```suggestion
	it('doesnt create an index for a canIdentify property if it does exist', async () => {
```

packages/tc-api-db-manager/index.js
```diff
@@ -4,7 +4,8 @@ const logger = require('@financial-times/n-logger').default;
 const schema = require('@financial-times/tc-schema-sdk');
 const dbConnection = require('./db-connection');
 
-const exclusion = (arr1, arr2) => arr1.filter(val => !arr2.includes(val));
+const exclusion = (arr1, arr2) =>
```
wheresrhys, line 7: I'm surprised this is needed. think there might be a mistake somewhere in flattening the nested lists of indexes/constraints into a single array.

We're on node12 now, so you could use [flatMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap) to simplify the horrible code from line 44 onwards

packages/tc-api-db-manager/__tests__/db-manager.spec.js
```diff
@@ -60,37 +78,91 @@ describe('creating db constraints', () => {
...
+		schema.getTypes.mockReturnValue([
+			{ name: 'Dog', properties: { nose: { canIdentify: true } } },
+		]);
+		await initConstraints();
+		expect(dbRun).not.toHaveBeenCalledWith('CREATE INDEX ON :Dog(nose)');
+
+		expect(dbRun).toHaveBeenCalledTimes(4);
+	});
+
+	it('doesnt create an index if a uniqueness constraint will be created', async () => {
```
i-like-robots, line 138: I don't really understand this description, is this interpretation correct?:

> does not create an index for fields with an existing unique constraint

Leigan0: Yep - your interpretation is much easier to understand so I will update that too. Thanks

packages/tc-api-db-manager/__tests__/db-manager.spec.js
```diff
@@ -35,21 +49,25 @@ describe('creating db constraints', () => {
...
 		]);
 		await initConstraints();
 		expect(dbRun).toHaveBeenCalledWith(
 			'CREATE CONSTRAINT ON (s:Dog) ASSERT s.tail IS UNIQUE',
 		);
+		expect(dbRun).toHaveBeenCalledTimes(5);
 	});
-
 	// ignore this test until we upgrade to enterprise adiition - community edition doesn't have exists()
 	it.skip("creates an existence constraint if it doesn't exist", async () => {
```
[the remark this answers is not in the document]
Leigan0: i'm not sure - I don't have the context on the upgrade - is this something that is going to happen?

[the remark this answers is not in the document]
wheresrhys: I think you could remove all the existence constraint stuff in a separate PR that's clearly named 'Remove existence constraint' so it's easy to find if we ever want to add it back in.

In the early days of the project it felt like an enterprise licence might be imminent, but it's turned out not to be the case - the api muddles on through without the benefits it would bring

PR #241, opened 2020-05-13
Fix submit on enter for relationship input 

Leigan0:
#### Bug Fix [ticket](https://financialtimes.atlassian.net/browse/RE-1781)

Create a child record at /ChildType/create

Go to the edit page for a MainType record and in the children field type the child's code - hopefully there will only be a single item in the autocomplete list

Press enter

Expected - Adds a row below the input contain the child record and clears the input and autocomplete

Actual behaviour - Adds a row below the input contain the child record and submits the entire form

## What?

-   When following the above steps the following error appeared in the console 

Screenshot 2020-05-13 at 09 45 19

-   This method `event.stopImmediatePropagation();` is not a function on keypress event so it stopped the prevent default 	
-   removed that line of code and this fixed the bug 

## Screenshots / Images

Before

### DIFF
--- a/packages/tc-ui/src/pages/edit/__tests__/e2e-create-type.cyp.js
+++ b/packages/tc-ui/src/pages/edit/__tests__/e2e-create-type.cyp.js
@@ -4,6 +4,7 @@ const {
 	promptText,
 } = require('../../../test-helpers/mainTypeData.json');
 const {
+	createType,
 	visitMainTypePage,
 	visitEditPage,
 	populateMinimumViableFields,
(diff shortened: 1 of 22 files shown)


packages/tc-ui/src/pages/edit/__tests__/e2e-create-type.cyp.js
```diff
@@ -166,5 +157,35 @@ describe('End-to-end - record creation', () => {
 					);
 			});
 		});
+		it('does not submit entire form when selection relationship with Enter', () => {
```
Leigan0, line 160:
```suggestion
		it('does not submit entire form when relationship input selection using Enter key press', () => {
```

Issue #245, opened 2020-05-15
Markdown formatting v yaml multi-line descriptions


GeoffThorpeFT:
**Describe the bug**
The formatting of multi-line Biz Ops schema description fields as markdown triggers the addition of extra blank lines; single line descriptions are fine.

**To Reproduce**
1) You need to be running treecreeper v0.5.10
2) Goto System/single-system-view/edit?#ownership-&-knowledge
3) Note the new line between `the` and `Group` within the description for `Delivered By Team`
4) Goto System/single-system-view/edit?#service-operability-review
5) Note how the `SOS traffic light` description is one nicely wrapped line of text.

**Expected behavior**
All field descriptions should appear as one nicely wrapped line unless they contain lines which whose markdown forces them on to new lines (such as the hyphen bullet point)

**Screenshots**
![image](https://user-images.githubusercontent.com/11228420/82027807-b5542580-968c-11ea-9e11-58f7ec5e3fb2.png)

![image](https://user-images.githubusercontent.com/11228420/82027853-c6049b80-968c-11ea-9643-cd2c60fdf7bd.png)

**Desktop (please complete the following information):**

-   OS: Windows
-   Browser: chrome
-   Version: 80

**Additional context**
The problem only occurs with treecreeper v0.5.10 since that version introduced descriptions with markdown formatting
v.0.5.10 is live in NPM but not built into any live apps at present.

GeoffThorpeFT:
We may need to balance the convenience of writing long descriptions as multi-line yaml with our wish to explicitly format some of those longer descriptions across multiple lines.
e.g. Where we just want to provide unformatted text we should use single line yaml for the description and let markdown/html split the line during rendering. However, when we want bullet points and bold we should use multiline yaml to ensure the markdown control characters are at the beginning of the lines.

GeoffThorpeFT:
We could also consider that this issue is caused by us writing multi-line yaml whose individual lines are too long (any therefore wrap triggering an unexpected new line).  In other words we could write a test to trigger failure when any line of a multi-line exceeds 80 characters.  Single line yaml can be as long as you want since they will wrap nicely.

GeoffThorpeFT:
After further investigation I note we are using both the `|` and `>` style of yaml to indicate multi-line text. The former introduced linefeeds while the later does not.  Ideally we would only use` |` for markdown where the linefeeds we necessary to force the correct positioning of bulltet points.

GeoffThorpeFT:
[...]

PR #247, opened 2020-05-15
Include a check for multiple descriptions which be corrupted by markdown

GeoffThorpeFT:
## Why?

-   Starts to fix: https://github.com/Financial-Times/treecreeper/issues/245
-   Multi-line descriptions are corrupted by markdown formatting due to their explicit new lines.

## What?

-   Check for single and multi-line descriptions
-   Single line descriptions can be of any length since markdown will generate html which wraps the text nicely
-   Multi-line descriptions can only be 80 characters per line since any more will trigger the inclusion of an extra new line (over and above the one present at the end of the line) during wrapping.
- The aim is to avoid wrapping lines within multi-line descriptions

This finds the descriptions that would cause the issue but doesn't fix them.

The first time the current `biz-ops-schema` is run against this version of the `tc-schema-validator` it will fail with the list of descriptions which need to be reformatted,

### DIFF
--- a/example-schema/types/MainType.yaml
+++ b/example-schema/types/MainType.yaml
@@ -23,6 +23,20 @@ properties:
     description: Word description.
     label: Word label
     useInSummary: true
+  hasLongDescription:
+    type: Word
+    description: This is a very long description of a schema item to prove that more than 80 characters can be used if a single line is used.
+    label: Word label
+    useInSummary: true
+  hasMutliLineDescription:
+    type: Word
+    description: |
+      These are many lines of decription each of which is of length 80 to prove length
+      These are many lines of decription each of which is of length 80 to prove length
+      These are many lines of decription each of which is of length 80 to prove length
+      These are many lines of decription each of which is of length 80 to prove sizes.
+    label: Word label
+    useInSummary: true
   # someStringList:
   #   type: Word
   #   hasMany: true
@@ -145,3 +159,4 @@ properties:
     type: AnotherEnum
     description: An enum that has something next to it.
     label: Multiple choice next to another thing
+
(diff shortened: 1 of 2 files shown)


GeoffThorpeFT:
I used 80 as it seems to be the point at which our origami screens wrap.  It was also very close to the manual splits we had already placed into the schema descriptions that used `|` for multiple lines.

wheresrhys:
> The aim is to avoid wrapping lines within multi-line descriptions

So is this just about making them look nicer in the github UI, or does it cause some error to have lines longer than 80?

wheresrhys:
Ah, now I see what the problem is. This fix doesn't feel very robust - descriptions can be used in other places and a change in styling would mean the layout breaks all over again.

Is this a good summary of the problem

> Descriptions with line breaks in them, when parsed as markdown, always (or sometimes?) have additional unwanted line breaks in the resulting html.

GeoffThorpeFT:
Lines look better in the GUI if they are less than 80 so this upgrades the schema tests run via treecreeper to highlight those descriptions which would look weird. 
In essence its looking for schema descriptions which have `|` instead of `>`

GeoffThorpeFT:
"Descriptions which have been defined with yaml pipe syntax, when parsed as markdown, always display their explicit line breaks even if the renderer has already wrapped the text."

wheresrhys:
what an annoying bug. I still think the focus on line length is not the right fix. Can the ui do a regex replace of new line followed by anything other than `` `#>-*(\d\(.|\)))``

GeoffThorpeFT:
IT really boils down to whether multi-line descriptions with explicit line breaks are allowed in the schema. If they are then the GUI should handle them exactly as the author intended.
If the only reason for multi-line is to introduce markdown formatting then we should probably have a 'isMarkdown' flag.

wheresrhys:
I have a task to write a 'how to review a biz-ops schema PR thoroughly' guide. 

What all this has highlighted is that linebreaks matter.

I think the advice to authors should be 'If writing a single para use `>` and break lines if you want. If writing markdown where you want to preserve line breaks, use `|` and be aware that lniebreaks will be preserved in the generated HTML.

> exactly as the author intended

The trouble is that this is hard to judge. We will never be able to tell if the author intends there to be a line break _exactly there_ or if they are adding a line break to make it look good in their text editor.

wheresrhys:
Have documented elesewhere the care that needs taking

PR #248, opened 2020-05-18
stop sending update events when noop deleted of relationship properties occur

wheresrhys:
## Why?
Since adding rich relationships to the model edits made in Biz Ops admin to systems have been causing excess update events in the kinesis stream.

After digging around quite a bit I discovered that the form was posting e.g `dependencies: [{code: 'abc', 'resiliencePatterns': null}]` when no relationship properties were added. The API was unable to tell that writing `null` to a previously undefined property had no effect, and instead thought this was a substantive change, causing an unnecessary write to the database and several change events in the kinesis stream.
## What?
- adds a test for this edge case
- explicitly handles the case where the existing property is not defined and a null value is sent, and treats this as 'no difference'

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "concurrently": "^5.0.0",
     "css-loader": "^3.2.1",
     "cypress": "^3.8.2",
-    "fetch-mock": "^9.6.0-alpha.2",
+    "fetch-mock": "^9.9.0",
     "isomorphic-fetch": "^2.2.1",
     "jest": "^24.8.0",
     "jest-junit": "^7.0.0",
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "concurrently": "^5.0.0",
     "css-loader": "^3.2.1",
     "cypress": "^3.8.2",
-    "fetch-mock": "^8.0.0",
+    "fetch-mock": "^9.6.0-alpha.2",
     "isomorphic-fetch": "^2.2.1",
     "jest": "^24.8.0",
     "jest-junit": "^7.0.0",
--- a/packages/tc-api-rest-handlers/lib/events/__tests__/events-integration.spec.js
+++ b/packages/tc-api-rest-handlers/lib/events/__tests__/events-integration.spec.js
@@ -226,7 +226,7 @@ describe('Rest events module integration', () => {
 			expectUpdateEvent(childType, childCode, ['isChildOf']);
 		});
 
-		it('should not send extra UPDATE events when making noop property removal edit to rich relationship', async () => {
+		it('should not send extra UPDATE events when removing rich relationship property that was already non-existent', async () => {
 			const [main, child] = await Promise.all([
 				createMainNode(),
 				createNode(childType, { code: childCode }),
--- a/packages/tc-api-rest-handlers/lib/events/__tests__/events-integration.spec.js
+++ b/packages/tc-api-rest-handlers/lib/events/__tests__/events-integration.spec.js
@@ -100,7 +100,6 @@ describe('Rest events module integration', () => {
 			expect(emitSpy).toHaveBeenCalledTimes(2);
 			expectCreateEvent(mainType, mainCode, [
 				'children',
-
 				'code',
 				'deprecatedChildren',
 				'someString',
@@ -225,6 +224,31 @@ describe('Rest events module integration', () => {
 				'someString',
 			]);
 			expectUpdateEvent(childType, childCode, ['isChildOf']);
+		});
+
+		it('should not send extra UPDATE events when making noop property removal edit to rich relationship', async () => {
+			const [main, child] = await Promise.all([
+				createMainNode(),
+				createNode(childType, { code: childCode }),
+			]);
+			await connectNodes(main, 'HAS_CURIOUS_CHILD', child);
+
+			const { status } = await patchHandler()(
+				getInput(
+					{
+						someString: 'some string',
+						curiousChild: {
+							code: childCode,
+							someString: null,
+						},
+					},
+					{ relationshipAction: 'replace', upsert: true },
+				),
+			);
+
+			expect(status).toBe(200);
+			expect(emitSpy).toHaveBeenCalledTimes(1);
+			expectUpdateEvent(mainType, mainCode, ['someString']);
 		});
 	});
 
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -38,7 +38,7 @@
     "concurrently": "^5.0.0",
     "css-loader": "^3.2.1",
     "cypress": "^3.8.2",
-    "fetch-mock": "^8.0.0",
+    "fetch-mock": "^9.9.0",
```
wheresrhys, line 41: also upgrades fetch-mock. Totally unrelated to the rest of the PR but I needed to test out a new release in a few projects and treecreeper is  a good 'n' complex one. May as well keep the upgrade now it's done

packages/tc-api-rest-handlers/lib/events/__tests__/events-integration.spec.js
```diff
@@ -226,6 +225,31 @@ describe('Rest events module integration', () => {
 			]);
 			expectUpdateEvent(childType, childCode, ['isChildOf']);
 		});
+
+		it('should not send extra UPDATE events when making noop property removal edit to rich relationship', async () => {
```
wheresrhys, line 229:
```suggestion
		it('should not send extra UPDATE events when removing rich relationship property that was already non-existent', async () => {
```

packages/tc-api-rest-handlers/lib/events/__tests__/events-integration.spec.js
```diff
@@ -226,6 +225,31 @@ describe('Rest events module integration', () => {
 			]);
 			expectUpdateEvent(childType, childCode, ['isChildOf']);
 		});
+
+		it('should not send extra UPDATE events when removing rich relationship property that was already non-existent', async () => {
+			const [main, child] = await Promise.all([
+				createMainNode(),
+				createNode(childType, { code: childCode }),
+			]);
```
wheresrhys, line 233: Create some fixture records - a `MainType` (`createMainNode() is a shorthand for `createNode('MainType', { code: mainCode })`... maybe that's a bit too DRY??

packages/tc-api-rest-handlers/lib/events/__tests__/events-integration.spec.js
```diff
@@ -226,6 +225,31 @@ describe('Rest events module integration', () => {
...
+			await connectNodes(main, 'HAS_CURIOUS_CHILD', child);
```
wheresrhys, line 234: create a relationship between the fixtures

packages/tc-api-rest-handlers/lib/events/__tests__/events-integration.spec.js
```diff
@@ -226,6 +225,31 @@ describe('Rest events module integration', () => {
...
+
+			const { status } = await patchHandler()(
+				getInput(
+					{
```
wheresrhys, line 238: getInput is a helper that return an object of the form `{code, type, body, metadata, query}`... i.e. the input for the rest handler functions

packages/tc-api-rest-handlers/lib/events/__tests__/events-integration.spec.js
```diff
@@ -226,6 +225,31 @@ describe('Rest events module integration', () => {
...
+
+		it('should not send extra UPDATE events when removing rich relationship property that was already non-existent', async () => {
+			const [main, child] = await Promise.all([
+				createMainNode(),
+				createNode(childType, { code: childCode }),
+			]);
+			await connectNodes(main, 'HAS_CURIOUS_CHILD', child);
+
+			const { status } = await patchHandler()(
+				getInput(
```
wheresrhys, line 237: getInput is a helper that return an object of the form {code, type, body, metadata, query}... i.e. the input for the rest handler functions

packages/tc-api-rest-handlers/lib/events/__tests__/events-integration.spec.js
```diff
@@ -226,6 +225,31 @@ describe('Rest events module integration', () => {
...
+					{
+						someString: 'some string',
+						curiousChild: {
+							code: childCode,
+							someString: null,
+						},
+					},
```
wheresrhys, line 244: The object above here is the `body` that gets sent. i.e., editing the `someString` property and basically writing the same relationship as already exists, including trying to delete a property on it which doesn't exist

packages/tc-api-rest-handlers/lib/events/__tests__/events-integration.spec.js
```diff
@@ -226,6 +225,31 @@ describe('Rest events module integration', () => {
...
+					{ relationshipAction: 'replace', upsert: true },
```
wheresrhys, line 245: the query params for the request

packages/tc-api-rest-handlers/lib/events/__tests__/events-integration.spec.js
```diff
@@ -226,6 +225,31 @@ describe('Rest events module integration', () => {
...
+				),
+			);
+
+			expect(status).toBe(200);
+			expect(emitSpy).toHaveBeenCalledTimes(1);
```
wheresrhys, line 250: this is a jest spy which keeps an eye on the event emitter built into the handler function

packages/tc-api-rest-handlers/lib/events/__tests__/events-integration.spec.js
```diff
@@ -226,6 +225,31 @@ describe('Rest events module integration', () => {
...
+			expectUpdateEvent(mainType, mainCode, ['someString']);
```
wheresrhys, line 251: this helper is basically saying `expect(emitSpy).calledWith({type: mainType, code: mainCode, updatedProperties: ['someString']})`

Is it too DRY to have a helper for this?? I dunno. Before writing helpers like this and the one I mention a few comments above the tests were equally unreadable due to the masses of repeated, verbose object literals. At least with the helpers once you've learned them the tests are easier to scan. But open to discussion. Perhaps what's needed is some 'before you read the tests, know this...' documentation

i-like-robots: It's this bit I don't understand - not because it's been extracted into a helper but because it's not clear why the input (below) would result in this outcome. Are there separate tests for the diff algorithm?

```
{
    someString: 'some string',
    curiousChild: {
        code: childCode,
        someString: null,
    },
}
```

PR #250, opened 2020-05-21
easier to get going with graphql explorer

wheresrhys:
## Why?

When getting Leanne and (IIRC) Kirsten onboarded I think they were both hit by problems connecting to the graphql explorer with no clear feedback on how to get unblocked.

## What?

- index page of the demo app links to the graphql explorer
- middleware always adds client id to api request made by the graphql explorer, so that the api does not error
- removed an old controller that was added to test some stuff out a while back but I don't think is useful any more

### DIFF
--- a/demo/api.js
+++ b/demo/api.js
@@ -21,11 +21,16 @@ app.use(
 	}),
 );
 
+// removes a little bit of hassle from manually testing the api
+app.use((req, res, next) => {
+	req.headers['client-id'] = 'treecreeper-demo';
+	next()
+})
+
 getApp({
 	treecreeperPath: '/api',
 	app,
 	graphqlMethods: ['post', 'get'],
-
 	documentStore: process.env.WITH_DOCSTORE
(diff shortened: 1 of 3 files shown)


demo/api.js
```diff
@@ -37,15 +44,14 @@ getApp({
...
-	editController,
-	viewController,
-	deleteController,
-	anotherController,
-} = require('./cms');
+const { editController, viewController, deleteController } = require('./cms');
 
 const parseBody = bodyParser.urlencoded({ limit: '8mb', extended: true });
-app.get('/lalala', anotherController);
+app.get('/', (req, res) => {
```
i-like-robots, line 50: If you want to code golf it then this is just fine, the browser can error correct the rest!

```<a href="/graphiql">GraphQL explorer.</a>```

wheresrhys: yeah - I was originally gonna add a lot more links but then decided against. Gotta love the 'most hostile development environment on the planet's ability to deal with bad input

PR #251, opened 2020-05-22
Fix #190

### LINKED ISSUE #190: Not able to save and render time field
This is a UI issue

To recreate:

```
curl -X PATCH http://local.ft.com:8888/api/rest/MainType/my-code -d '{"someTime": "12:15:32"}' -H 'client-user-id: rhys' -H 'content-type: application/json' 
```

creates a record with a value in the time field.

The view page (http://local.in.ft.com:8888/MainType/my-code) errors with this error:

image

On the edit page (http://local.in.ft.com:8888/MainType/my-code/edit) the UI does not error, but renders no value, despite the data attribute showing the value is passed into the component
image

wheresrhys:
## Why?

The UI did not cope with fields of type Time #190 

## What?
- added more hacks to the value parsing. Ultimately the behaviour around how to handle timezones is very hacky, but without a real case it's hard to know what behaviour to implement. For now the focus is on 'just don't error' so that times that don't use time zones do at least work
- cypress tests for time, but also added tests for datetime and date to make sure values that are saved can then be redisplayed successfully in the edit component and resaved i.e user input -> db  -> parse back to form data -> save

### DIFF
--- a/packages/tc-ui/src/primitives/temporal/server.jsx
+++ b/packages/tc-ui/src/primitives/temporal/server.jsx
@@ -17,7 +17,7 @@ const formatTemporal = (value, type) => {
 	}
 
 	if (type === 'Time') {
-		return formatValue(`2020-01-15T${value}`, 'h:mm:ss a')
+		return formatValue(`2020-01-15T${value}`, 'h:mm:ss a');
 	}
 };
 
@@ -31,7 +31,7 @@ const convertValueForHTMLInput = (wrappedValue, type) => {
 	// the input doesn't error
 	// Should probably store the timestamps without any timezone info for
 	// consistency, but that's a bug to fix in future
-	if (type === 'Time') return value.split(/Z|\+|\-/)[0];
+	if (type === 'Time') return value.split(/Z|\+|-/)[0];
 	const date = new Date(value).toISOString();
 	return type === 'DateTime' ? date.split('Z')[0] : date.split('T')[0];
 };
@@ -62,7 +62,9 @@ const EditTemporal = ({
 			<input
 				name={name}
 				id={`id-${propertyName}`}
-				type={type === 'DateTime' ? 'datetime-local' : type.toLowerCase()}
+				type={
+					type === 'DateTime' ? 'datetime-local' : type.toLowerCase()
+				}
 				step={type === 'Time' ? 1 : null}
 				value={convertValueForHTMLInput(value, type)}
 				required={required ? 'required' : null}
--- a/packages/tc-ui/src/primitives/temporal/server.jsx
+++ b/packages/tc-ui/src/primitives/temporal/server.jsx
@@ -7,12 +7,7 @@ const { WrappedEditComponent } = require('../../lib/components/input-wrapper');
 const formatValue = (value, formatString) =>
 	value ? format(parseISO(value), formatString) : null;
 
-// TODO this probably errors - need to write comprehensive tests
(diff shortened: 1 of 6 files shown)


packages/tc-ui/src/primitives/temporal/server.jsx
```diff
@@ -25,19 +17,22 @@ const formatDateTime = (value, type) => {
...
-		return formatTime(timeInput);
+		return formatValue(`2020-01-15T${value}`, 'h:mm:ss a');
 	}
 };
 
 const convertValueForHTMLInput = (wrappedValue, type) => {
 	if (!(wrappedValue && wrappedValue.formatted)) return null;
 	const value = wrappedValue.formatted;
-	if (type === 'Time') return value;
+
```
Leigan0, line 27: Just a question around this really, but if we have no use case might not be an answer. But I guess if people input through the UI we don't add a timezone? so if someone is adding a timezone  through the API then it would suggest it would be important to the user? so wouldn't it make more sense to add a default timezone based on user location or something  rather than strip the timezone?

packages/tc-ui/src/primitives/temporal/server.jsx
```diff
@@ -25,19 +17,22 @@ const formatDateTime = (value, type) => {
 	}
 
 	if (type === 'Time') {
-		const timeInput = value;
-		return formatTime(timeInput);
+		return formatValue(`2020-01-15T${value}`, 'h:mm:ss a');
```
Leigan0, line 20: This is a bit confusing - if you used moment you can just do 
`var time = moment('12:15:33', 'HH:mm:ss a');` without having to add a date and convert into ISOstring but not sure its worth the change

packages/tc-ui/src/primitives/temporal/server.jsx
wheresrhys: Trouble is that the `type="time"` input doesn't support displaying a time with a timezone. We didn't realise this though until after the API already implemented saving times with a default timezone (which I don't think was deliberate, it just sorta happened).

So in the longer term there's 1 of 2 paths to follow (depending on what our users actually want)
1. rewrite the API to never save timezones
2. rewrite the UI to support editing timezones

I don't really want to do either until we know what's required

packages/tc-ui/src/primitives/temporal/server.jsx
wheresrhys: I agree moment is a bit simpler, but it is notorious for being really big, which is why I avoid it.

Although I just checked out their website and it looks like the latest version is much smaller.

Leigan0: Yea I saw that when I compared the two which was surprising. Although I guess this works for now so..?

packages/tc-ui/src/primitives/temporal/server.jsx
Leigan0: fair enough - that makes sense. 

What would the plan be to find out? or would you wait until it was mentioned in feedback from a user?

packages/tc-ui/src/primitives/temporal/server.jsx
wheresrhys: I had a go at using moment. For now I'm gonna pass as I couldn't get it to work and don't have the time to explore more. I'll leave a comment in the code to explain why we add eth date though as could benefit from an explanation

PR #255, opened 2020-06-15
Add fieldset description

Leigan0:
## Why?

-  [ticket](https://financialtimes.atlassian.net/browse/RE-1844)

It’s possible to add a description for a fieldset in the schema but this isn’t used anywhere

We have made use of these within risk model 

## What?

- Created a new type in the demo schema `FieldsetType`
- add code to use the fieldset description on the edit and view pages
- added tests for create/view/edit  views to confirm fieldsets and description displays as expected. 

n/a

## Screenshots / Images

Screenshot 2020-06-15 at 16 28 56

Screenshot 2020-06-15 at 16 29 09

### DIFF
--- a/packages/tc-ui/src/pages/view/template.jsx
+++ b/packages/tc-ui/src/pages/view/template.jsx
@@ -96,7 +96,7 @@ const View = props => {
 					</div>
 					<div className="o-layout-typography">
 						{Object.entries(schema.fieldsets).map(
-							([name, { heading, properties }]) => (
+							([name, { heading, description, properties }]) => (
 								<section
 									className={`fieldset-treecreeper fieldset-${name}`}
 								>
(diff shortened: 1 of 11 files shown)


packages/tc-ui/src/pages/view/template.jsx
```diff
@@ -96,24 +96,23 @@ const View = props => {
 					</div>
 					<div className="o-layout-typography">
 						{Object.entries(schema.fieldsets).map(
-							([name, { heading, properties }]) => (
-								<section
-									className={`fieldset-treecreeper fieldset-${name}`}
+							([name, { heading, description, properties }]) => (
+								<Fieldset
```
wheresrhys, line 100: Is there a way you can force the Fieldset component to use the `<section>` html tag on the view page? `<fieldset>` shouldn't really be used outside of forms, for accessibility reasons

Leigan0: Ahh ok I didn't know that. Good to know - so I have updated it to make it conditional on `includeEditLink` as that will only show on the view. Do you think it would benefit from being based on a more explicit prop - like `displayOnForm` or something?

wheresrhys: Maybe something like `isInEditMode`, which acts as the trigger for all the differences in output between edit and view

PR #257, opened 2020-06-17
prevent blocking of empty strings by validation patterns

wheresrhys:
## Why?

@nl-ria  found a bug whereby empty strings in fields that used a vaildation pattern were being rejected. This is because the pattern was running against all strings, even empty ones.

## What?
Adds a test and coerces empty strings to null early. Empty strings are already treated as (or are supposed to be, at least) null when writing to the database, so this shifts that coercion earlier in order to skip validation.

Similar thing needs to happen for properties on relationships - i will add an issue to capture that requirement, but won't tackle it here because it'll get way more complicated and does not actually apply to any of our Biz Ops data at the moment - we have no string-validtead properties written onto relationships

I do now wonder if there's been a bug for a while that means we've been saving empty strings to the DB - I will do some kind of audit on the data

### DIFF
--- a/packages/tc-api-rest-handlers/lib/validation.js
+++ b/packages/tc-api-rest-handlers/lib/validation.js
@@ -94,9 +94,22 @@ const validateMetadata = ({
 	}
 };
 
+const coerceEmptyStringsToNull = ({ body }) => {
+	Object.entries(body).forEach(([propName, value]) => {
+		if (value === '') {
+			body[propName] = null;
+		}
+	});
+};
+
 const validateInput = input => {
 	validateParams(input);
 	if (input.body) {
+		// TODO need to do something similar for relationship properties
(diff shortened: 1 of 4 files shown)


packages/tc-api-rest-handlers/lib/validation.js
```diff
@@ -94,9 +94,22 @@ const validateMetadata = ({
...
+			body[propName] = null;
+		}
+	});
+};
+
 const validateInput = input => {
 	validateParams(input);
 	if (input.body) {
+		// TODO need to do something similar for relationship properties
+		// and consolidate with eth isNull checks in diff-properties.js
```
nl-ria, line 109:
```suggestion
		// and consolidate with the isNull checks in diff-properties.js
```

PR #261, opened 2020-06-23
fix empty input for int type should defaulting to 0 

Leigan0:
### Bug Fix

-   [ticket](https://financialtimes.atlassian.net/secure/RapidBoard.jspa?rapidView=1169&projectKey=RE&modal=detail&selectedIssue=RE-1852)
-   if you save the form for a risk with empty strings in the number fields, they get saved as zeros. 
expected behaviour - save as null

## What?

-   Updated number parser to check for empty string, and if so return null entry to be sent to treecreeper API 
-   End user expected behaviour of this is that these fields will not display on success screen. Added test to replicate this. 

Mention here sections of code which you would like reviewers to pay extra attention to, for example:

_I have made a few assumptions here - there may be edge cases or further places this needs updating_  

## Screenshots / Images

Screenshot 2020-06-23 at 11 27 49

Screenshot 2020-06-23 at 11 31 21

### DIFF
--- a/packages/tc-ui/src/primitives/number/server.jsx
+++ b/packages/tc-ui/src/primitives/number/server.jsx
@@ -4,6 +4,9 @@ module.exports = {
 	name: 'Number',
 	EditComponent: text.EditComponent,
 	ViewComponent: text.ViewComponent,
-	parser: value => (Number.isNaN(Number(value)) ? value : Number(value)),
+	parser: value => {
+		if (value === '') return null;
+		return Number.isNaN(Number(value)) ? value : Number(value);
+	},
 	hasValue: value => value || value === 0,
 };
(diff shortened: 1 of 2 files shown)


packages/tc-ui/src/primitives/number/server.jsx
```diff
@@ -4,6 +4,9 @@ module.exports = {
 	name: 'Number',
 	EditComponent: text.EditComponent,
 	ViewComponent: text.ViewComponent,
-	parser: value => (Number.isNaN(Number(value)) ? value : Number(value)),
+	parser: value => {
+		if (value === '') return null;
+		return Number.isNaN(Number(value)) ? value : Number(value);
+	},
 	hasValue: value => value || value === 0,
```
wheresrhys, line 11: does this line need to change too? A related bug, I guess the current implementation means the field shows up on the view page showing '0' even when it was originally empty.  Not 100% sure of that, but worth checking and fixing if necessary

wheresrhys: https://github.com/Financial-Times/treecreeper/blob/300c6a1214758ee53f0fb25d00e904e7c6f84019/packages/tc-ui/src/lib/components/structure.jsx#L67 is why I think it might cause this bad behaviour

wheresrhys: actually, I don't think this is a problem - ignore me

PR #262, opened 2020-06-24
Truncate strings

lucas42:
## Why?

@wheresrhys reported that https://biz-ops-test.in.ft.com/System/sso-subscription-service-my-licence/edit was returning a 503.  Based on the logs in splunk, this was caused by a 413 entity too large error.  According to [the AWS docs](https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html), the limit payload limit for lambdas is 6MB
For this particular system (at least based on data in test) we're returning a payload of 11.4MB
This looks to be caused by each base64 encoded image included in troubleshooting section being output in the payload 20 times.

## What?

Rather than outputting the entire biz-ops record as a json-encoded attribute for every component, output the same structure, but truncate any large strings at the top level

This isn't the nicest solution, but as I'm not familiar with all the interdependencies of the various treecreeper bits, I wanted to keep any changes to the tc/biz-ops interface as minimal as possible to avoid breaking anything.

I the longer term, I think we should consider removing the `entireRecord` attribute (or at least only include it once per page, rather than per component)

## Screenshots / Images

Before
11.4MB

After
610kB

### DIFF
--- a/packages/tc-ui/src/pages/edit/template.jsx
+++ b/packages/tc-ui/src/pages/edit/template.jsx
@@ -31,10 +31,23 @@ const PropertyInputs = ({ fields, data, type, assignComponent, hasError }) => {
 				? data[`${propertyName}_rel`] || data[propertyName]
 				: data[propertyName];
 
+			/**
+			 * Sending the entire record to the client multiple times can increase the payload immensely
+			 * However some components rely on this (eg "decommissioned" functionality)
+			 * As a workaround for now, truncate long text fields
+			 * TODO: think of a better way to be selective about what we send
+			 */
+			const clientSideRecord = {};
+			Object.entries(data).forEach(([key, value]) => {
+				clientSideRecord[key] =
+					typeof value !== 'string' || value.length < 25
+						? value
+						: `${value.substring(0, 24)}…`;
+			});
 			const viewModel = {
 				hasError,
 				parentCode: data.code,
-				entireRecord: data,
+				entireRecord: clientSideRecord,
 				propertyName,
 				value: getValue(propDef, itemValue),
 				dataType: propDef.type,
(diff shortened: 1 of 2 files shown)


wheresrhys:
Isn't this quite dangerous. If a user is on the edit page, the large fields will contain truncated content. If they click save then the truncated content will be saved to Biz Ops.

There are other reasons to move biz ops admin to heroku, and this provides another compelling one cc @i-like-robots

wheresrhys:
ah no, I see why this might be safe, because large documents cannot be saved on relationships (which is where the entire serialised object is used) .Will review properly after lunch

lucas42:
Yes, also I'm only truncating strings at the top level of the record - relationship data will be nested deeper, so unaffected.

packages/tc-ui/src/pages/edit/template.jsx
```diff
@@ -31,10 +31,23 @@ const PropertyInputs = ({ fields, data, type, assignComponent, hasError }) => {
...
 
+			/**
+			 * Sending the entire record to the client multiple times can increase the payload immensely
+			 * However some components rely on this (eg "decommissioned" functionality)
+			 * As a workaround for now, truncate long text fields
+			 * TODO: think of a better way to be selective about what we send
+			 */
+			const clientSideRecord = {};
+			Object.entries(data).forEach(([key, value]) => {
+				clientSideRecord[key] =
```
wheresrhys, line 42: a nicer test would be `properties[key].type !== 'Document'`

lucas42: I though a schema driven approach would be good, but didn't really understand the different between `fields`, `properties` and `propertyfields`.

You sure it shouldn't be `fields[key].type`?

PR #263, opened 2020-06-26
avoid outputting cypher properties in the ui

wheresrhys:
## Why?

The UI breaks when properties make use of the powerful @cypher directive. There's another PR to sketch out what being able to display these properties in the UI might look like https://github.com/Financial-Times/treecreeper/pull/256...

But I realised that is quite a big job, and using @cypher is useful now (in particular for distilling risk blast radiuses down into summaries, or extracting other useful summaries from the new more ophisticated modelling we're doing of prosucts, capabilities, Fastly, Hosts etc).

So this PR implements the relatively small chaneg of suppressing any attempt to render these fields in the admin UI, saving them for the GraphQL api alone. This means we can define them and explore using them, then take on the UI job as and when it feels useful to do so

## What?

- Adds a new `includeSyntheticFields` property to `getType`, defaulting to `true` (for backwards compat)
- This is then used when types are accessed within tc-ui
- updates some schema validation and schema sdk tests

## TODO

API validation needs updating so that its' impossible to accidentally write a cypher property - this should error informatively:

```bash
curl -X PATCH 'http://local.in.ft.com:8888/api/rest/MainType/bob?upsert=true&relationshipAction=merge' --data '{"cypherThing": "tony"}' -H 'client-id: rhys' -X 'content-type: application/json'
```

*update - there is a test for this and it _should_ 400, but not doing so for some reason*
*update 2 - false alarm, the api call was malformed. The validation is good as it is*

### DIFF
--- a/packages/tc-schema-sdk/README.md
+++ b/packages/tc-schema-sdk/README.md
@@ -104,6 +104,7 @@ The full object structure returned by getType() can been seen [here](GETTYPE.md)
 -   `withRelationships` [default: `true`]: Include the relationships for the type, expressed as graphql property definitions.
 -   `groupProperties` [default: `false`]: Each property may have a `fieldset` attribute. Setting `groupProperties: true` removes the `properties` object from the data, and replaces it with `fieldsets`, where all properties are then grouped by fieldset
 -   `includeMetaFields` [default: `false`]: Determines whether to include metadatafields (prefixed with `_`) in the schema object returned
+-   `includeMetaFields` [default: `true`]: Determines whether to include synthetic fields (those using a custom cypher statement) in the schema object returned
 -   `useMinimumViableRecord` [default: `false`]: If `groupProperties` is `true`, this will put any fields defined as being part of the minimum viable record (see [model spec](MODEL_SPECIFICATION.md#types)) together in a single fieldset
 
 ### getTypes(options)
--- a/packages/tc-schema-sdk/data-accessors/__tests__/type.spec.js
+++ b/packages/tc-schema-sdk/data-accessors/__tests__/type.spec.js
@@ -89,6 +89,27 @@ describe('get-type', () => {
 		});
 	});
 
+	it('can exclude meta properties', () => {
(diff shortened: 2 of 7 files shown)


packages/tc-schema-sdk/README.md
```diff
@@ -104,6 +104,7 @@ The full object structure returned by getType() can been seen [here](GETTYPE.md)
 -   `withRelationships` [default: `true`]: Include the relationships for the type, expressed as graphql property definitions.
 -   `groupProperties` [default: `false`]: Each property may have a `fieldset` attribute. Setting `groupProperties: true` removes the `properties` object from the data, and replaces it with `fieldsets`, where all properties are then grouped by fieldset
 -   `includeMetaFields` [default: `false`]: Determines whether to include metadatafields (prefixed with `_`) in the schema object returned
+-   `includeMetaFields` [default: `true`]: Determines whether to include synthetic fields (those using a custom cypher statement) in the schema object returned
```
wheresrhys, line 107:
```suggestion
-   `includeSyntheticFields` [default: `true`]: Determines whether to include synthetic fields (those using a custom cypher statement) in the schema object returned
```

Leigan0, line 107: should this be `includeSyntheticFields` ?
```suggestion
-   `includeSyntheticFields` [default: `true`]: Determines whether to include synthetic fields (those using a custom cypher statement) in the schema object returned
```

packages/tc-schema-sdk/data-accessors/__tests__/type.spec.js
```diff
@@ -89,6 +89,27 @@ describe('get-type', () => {
...
+			{
+				name: 'Type1',
+				description: 'I am Type1',
+				properties: {
+					property1: {
+						type: 'String',
+					},
+				},
+			},
+			{ options: { includeMetaFields: false } },
```
i-like-robots, line 104: synthetic? Otherwise this may indicate the test doesn't do what it is intended!

PR #272, opened 2020-07-20
loosen tests to allow integer codes

wheresrhys:
## Why?

Response bot uses incrementing integers as identifiers fro incidents. However, when importing data using `Int` as the type for the code of Incident records is not permitted, leading to some jhacks around adding a string pattern that won't be used, or using a String type and writing `"123"` rather than `123`

## What?
Changes the tests to be intelligent enough to have relaxed expectations around patterns if an Int type is used

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ cypress/screenshots/
 cypress/videos/
 cypress/fixtures/
 dist/
+.env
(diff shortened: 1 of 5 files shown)


wheresrhys:
There's a real problem with the build pipeline, so might be a long time getting this released

wheresrhys:
Tried manually creating records with integer codes and discovered there are lots of other code changes needed to 
a) avoid coercing code properties to string
b) avoid treating codes provided as url params as strings

Too much work for too little benefit

PR #275, opened 2020-07-29
Allow custom resolver to co-exist with document resolver

lucas42:
## Why?

Currently if you add a customer resolve to a Type which has any document fields, the document resolvers get removed.

## What?

Using lodash's merge function to deep merge the resolvers objects, rather than the shallow Object.assign.  This enables both sets of resolvers to exist side-by-side.

### DIFF
--- a/packages/tc-api-graphql/__tests__/graphql.spec.js
+++ b/packages/tc-api-graphql/__tests__/graphql.spec.js
@@ -318,6 +318,11 @@ describe('graphql', () => {
 						}),
 					},
 				},
+				documentStore: {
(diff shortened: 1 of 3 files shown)


lucas42:
Oh, I saw lodash was already being used in the project, so thought it was better to re-use that than pull in another dependency.  Have you got any recommendations of good alternatives?

Also, any idea how to get the tests in circleCI to work?  They keep complaining of a DB init failure.  (I've got them passing locally).

lucas42:
I've merged in your commit that fixes the tests #277 and raised an issue around the use of lodash #278

PR #279, opened 2020-07-29
attempt to fix bug in use of neo4j-driver

wheresrhys:
## Why?

https://github.com/neo4j/neo4j-javascript-driver/issues/606

## What?

Being a bit more careful when awaiting db results in the code path taken during tests (which do not apply a timeout).


lucas42:
Looks like that hasn't fixed it
image
The stacktrace suggests the error is happening inside the `runner()` function, so I don't think awaiting on it will be enough.

PR #308, opened 2020-10-08
Simple relationship

wheresrhys:
## Why?

Replacing the old relationship component with the rich relationship one was, in hindsight, an error. It means reusing the components is harder. We updated runbokos to use the components and it's meant we have to add all the e.g. `dependencies_rel` as well as `dependencies` to the graphql query - an ugly mess.

Having a simple relationship component that can consume `dependencies` as well as s rich relationship component that can consume `dependencies_rel` gives more flexibility of reuse.

This will also allow us to start surfacing the synthetic properties defined using `@cypher` in biz ops admin (but might avoid doing that at first while we decide how to make the UI clearer around their non-editability)

## What?

- Copies the old relationship component in from a much earlier version of the repo
- Renames the relationship component to RichRelationship
- Also made the relationship annotator interface a bit more powerful - given access to property name and relationship properties, which will aid with making better use of annotated relationships (there have been a few issues raised it'd help with over the past few months)

TODO
- [x] cypress tests for Relationship
- [x] dedupe styles
- [x] port any bugfixes in the RichRelationship component back into Relationship
- [x] avoid displaying `@cypher` properties for now
- [x] release as a beta
- [x] try using in biz-ops-admin
- [ ] try using in biz-ops-runbooks, mainly using SimpleRelationship instead of Relationship in most places
- [ ] release as patch (not minor as we are < v0 which means minors ~= majors)
- [ ] use the patch version in biz-ops-admin and runbooks


wheresrhys:
A thought to put the brakes on this

Say a relationship (without properties on it) has a custom component defined in biz ops admin. If I add properties to the relationship in the schema this means the custom component will now be augmenting the wrong base component (Relationship rather than RichRelationship). That could be disastrous.

In hindsight, maybe the simple relationship component should only exist as a view utility for viewing relationships defined using `@cypher`, or for reuse in other apps

wheresrhys:
Actually it wouldn't break, because relationship annotations always use the method to inject an additional component. _But_ I was introducing so many hacks to get the new component to work in the edit page that it seems wise to hold off on that until such time as we tidy the entire package

wheresrhys:
Made a last minute adjustment to not change the name of the existing component, but to call the new component `SimpleRelationship`. Longer term, Relationship and RichRelationship probably makes more sense, but this avoids a breaking change and a painful roll out

packages/tc-ui/src/pages/edit/browser.js
```diff
@@ -21,7 +21,7 @@ const initDocumentEditors = () => {
 const initRelationshipSelectors = entireRecord => {
 	[
 		...document.querySelectorAll(
-			'[data-component="relationship-picker"]:not([data-disabled])',
+			'[data-component="relationship"]:not([data-disabled])',
 		),
 	].forEach(container => attachRelationshipPicker(container, entireRecord));
 };
```
i-like-robots, line 27: No need to spread a nodelist into an array, `NodeList.prototype.forEach` has been a thing for a long time!

packages/tc-ui/src/primitives/boolean/server.jsx
```diff
@@ -52,13 +52,13 @@ const Checkbox = ({
...
 	const {
 		propertyName,
-		value,
 		disabled,
 		isNested,
 		parentType,
 		parentCode,
 		onChange,
 	} = props;
+	const value = typeof props.value === 'boolean' ? props.value : '';
```
i-like-robots, line 61: I feel like this is a breaking change... are there any potential risks with this?

packages/tc-ui/src/primitives/multiple-choice/server.jsx
```diff
@@ -87,15 +87,15 @@ const Checkbox = ({
...
-		options,
 		disabled,
 		isNested,
 		parentCode,
 		onChange,
 		type,
 	} = props;
 	const name = !isNested ? propertyName : '';
+	const options = Object.keys(getEnums()[type]);
+	const value = props.value ? props.value.sort() : [];
```
i-like-robots, line 98: Should the array be cloned before sorting as this will sort in place rather than create a new array?

packages/tc-ui/src/primitives/simple-relationship/server.jsx
```diff
@@ -0,0 +1,75 @@
...
+const { SelectedRelationship } = require('./lib/selected-relationship');
+
+const RelationshipPickerContainer = props => (
+	<div>
+		<RelationshipPicker
+			{...props}
+			SelectedRelationship={SelectedRelationship}
+			componentName="simple-relationship-picker"
+		/>
+	</div>
```
i-like-robots, line 14: No need for additional `<div>`

packages/tc-ui/src/primitives/simple-relationship/server.jsx
```diff
@@ -0,0 +1,75 @@
...
+			return null;
+		}
+		value = JSON.parse(value);
+		// TODO use hasValue
+		if (!value) {
+			return null;
+		}
+		return Array.isArray(value)
+			? value.map(({ code }) => code)
+			: value.code;
```
i-like-robots, line 52: Should this also be an array so that the parser only returns one type? i.e. `[ value.code ]`

packages/tc-ui/src/primitives/simple-relationship/lib/view-relationship.jsx
```diff
@@ -0,0 +1,61 @@
+const React = require('react');
+const { getType } = require('@financial-times/tc-schema-sdk');
+const { LinkToRecord } = require('../../../lib/components/structure');
+
+let RelationshipAnnotator;
```
i-like-robots, line 5: This reference scares me a bit... I assume a record may have different types of relationship so this may be toggled multiple times to render a page. If rendering ever becomes async - or if async logic added between the `setRelationshipAnnotator()` call and rendering - this could cause some wild problems.

packages/tc-ui/src/primitives/boolean/server.jsx
wheresrhys: It's a port of this https://github.com/Financial-Times/treecreeper/pull/308/files#diff-0a84d66ba70a9ea00b0d958dee53c7b90a9d3cda89c5d7352a6d94c17719a104L6

I've basically distributed that file around the primitive components

packages/tc-ui/src/primitives/multiple-choice/server.jsx
wheresrhys: Yes... but I don't actually know why that is even sorted (again, a port from the old get-value file)

packages/tc-ui/src/primitives/simple-relationship/server.jsx
wheresrhys: Needed for styling because flexbox I think. I can leave a comment

packages/tc-ui/src/primitives/simple-relationship/server.jsx
wheresrhys: no, because for hasMany fields the values are `[]` or `[some, values]`, whereas for non-hasMany they are a code (string) or null. 

In time the logic here (and in many other places) could do with branching once based on a clear criteria such as `if (hasMany) { ... }`, but not aiming to do that in this PR

packages/tc-ui/src/primitives/simple-relationship/lib/view-relationship.jsx
wheresrhys: Can you add a comment to the 'do tc-ui properly' issue. Would be good to use that as a place to collect things to address. For now though, this is completely server side rendered and the risk is low

packages/tc-ui/src/pages/edit/browser.js
wheresrhys: This is done quite a bit and not very well tested. Don't want to either a) tidy just this one b) tidy them all but introduce a mistake... so gonna leave it for now

packages/tc-ui/src/primitives/multiple-choice/server.jsx
wheresrhys: I _think_ it was probably copied from the relationship component (where order matters as we want to display them alphabetically ,whereas here they are displayed in the order the enum defines)

PR #327, opened 2021-01-06
update example structure to provide custome from and to nested node names

wheresrhys:
## Why?

https://financialtimes.atlassian.net/browse/RE-2275

## What?

Provides data points in the schema yaml to enable building out the new GraphQL schema

### DIFF
--- a/example-schema/relationships/CuriousSibling.yaml
+++ b/example-schema/relationships/CuriousSibling.yaml
@@ -1,9 +1,11 @@
 name: CuriousSibling
 from:
   type: MainType
+  otherNodeName: youngerSibling
   hasMany: true
 to:
   type: MainType
+  otherNodeName: olderSibling
   hasMany: true
 relationship: HAS_YOUNGER_ANNOTATED_SIBLING
 properties:
--- a/example-schema/types/MainType.yaml
+++ b/example-schema/types/MainType.yaml
@@ -102,6 +102,7 @@ properties:
   youngerSiblings:
     relationship: HAS_YOUNGER_SIBLING
     type: MainType
+    otherNodeName: youngerSibling
     description: Younger siblings description.
     label: Younger siblings label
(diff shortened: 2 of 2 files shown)


example-schema/relationships/CuriousSibling.yaml
```diff
@@ -1,9 +1,11 @@
 name: CuriousSibling
 from:
   type: MainType
+  otherNodeName: youngerSibling
```
GeoffThorpeFT, line 4: `otherNodeName` doesn't feel like the right name for this concept as its only applicable when the `from` and `to` types are the same.  Something like `selfReferenceAlias` would be better but thats still not right.

example-schema/types/MainType.yaml
```diff
@@ -104,13 +104,15 @@ properties:
     type: MainType
     description: Younger siblings description.
     label: Younger siblings label
+    otherNodeName: youngerSibling
```
GeoffThorpeFT, line 107: As mentioned above (in the relationships yaml) the name of the property doesnt highlight that its naming the target of a self-reference and is therefore not required for every relationship.

GeoffThorpeFT: It might also help if this property was specified directly after the `type` property to highlight that its effect is to rename that type (yes, I know the order is not important syntactically but it could improve readability)

example-schema/relationships/CuriousSibling.yaml
wheresrhys: I looked at the GrandStack docs and maybe `...FieldName` would be good

`unambiguousFieldName` is the reason for setting it, but that's a bit off

either way, we can use it as a working title and do a big find and replace if we come up with better

wheresrhys:
as discussed elsewhere, changes by neo4j to the entire neo4j-graphql direction mean this is no longer needed

PR #329, opened 2021-01-13
Neo4j graphql upgrade sdk

GeoffThorpeFT:
## Why?

-   The SDK needs to support the use of aliases when a relationship references a parent/child/sibling of the same type.

## What?

-   Extend the `getRelationship `function to add `otherNodeName` to the output of every simple relationship
-   Create a new `formatRichRelationship` function to allow getRelationship function to add `otherNodeName` to the output of every rich relationship
-   Extend the tests for get-relationship to expect `otherNodeName` to always be present

WIP
`otherNodeName` will always be output by getRelationship even when it was not explicitly supplied in the yaml; it that case it will be set to the parent type.

### DIFF
--- a/example-schema/relationships/CuriousSibling.yaml
+++ b/example-schema/relationships/CuriousSibling.yaml
@@ -1,9 +1,11 @@
 name: CuriousSibling
 from:
   type: MainType
+  otherNodeName: youngerSibling
   hasMany: true
 to:
   type: MainType
+  otherNodeName: olderSibling
   hasMany: true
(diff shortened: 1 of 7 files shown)


wheresrhys:
Could you open a fresh PR onto my branch, because otherwise approval of this implies merging to master, but we want to merge all the partial changes onto the other branch first

GeoffThorpeFT:
Sorry, thought I created the sub branch - must have mucked up the command :(

GeoffThorpeFT:
Rebase into this PR: https://github.com/Financial-Times/treecreeper/pull/330

PR #330, opened 2021-01-13
Neo4j graphql upgrade sdk

GeoffThorpeFT:
## Why?

-   The SDK needs to support the use of aliases when a relationship references a parent/child/sibling of the same type.

## What?

-   Extend the `getRelationship `function to add `otherNodeName` to the output of every simple relationship
-   Create a new `formatRichRelationship` function to allow getRelationship function to add `otherNodeName` to the output of every rich relationship
-   Extend the tests for get-relationship to expect `otherNodeName` to always be present

WIP
`otherNodeName` will always be output by getRelationship even when it was not explicitly supplied in the yaml; it that case it will be set to the parent type.

### DIFF
--- a/packages/tc-schema-sdk/data-accessors/relationship-type.js
+++ b/packages/tc-schema-sdk/data-accessors/relationship-type.js
@@ -62,14 +62,17 @@ const formatSimpleRelationship = (
 	};
 };
 
-const formatRichRelationship = richRelationshipDefinition => {
+const setOtherNodeNameWhenAbsent = richRelationshipDefinition => {
 	const { from, to } = richRelationshipDefinition;
-	const enriched = {
+	if (from.otherNodeName && to.otherNodeName) {
+		return richRelationshipDefinition;
+	}
+
+	return {
 		...richRelationshipDefinition,
-		from: {otherNodeName: from.type, ...from },
-		to: {otherNodeName: to.type, ...to },
+		from: { otherNodeName: from.type, ...from },
+		to: { otherNodeName: to.type, ...to },
 	};
-	return enriched;
 };
 
 const getTypeProperty = (rootType, propertyName, rawData) => {
@@ -119,7 +122,7 @@ const getRelationshipTypeFromRawData = (rootType, propertyName, rawData) => {
 	);
 
 	if (richRelationshipDefinition) {
-		return formatRichRelationship(richRelationshipDefinition);
+		return setOtherNodeNameWhenAbsent(richRelationshipDefinition);
 	}
 
 	throw new TreecreeperUserError(
--- a/packages/tc-schema-sdk/data-accessors/relationship-type.js
+++ b/packages/tc-schema-sdk/data-accessors/relationship-type.js
@@ -66,11 +66,9 @@ const formatRichRelationship = richRelationshipDefinition => {
 	const { from } = richRelationshipDefinition;
 	const { to } = richRelationshipDefinition;
 	const enriched = {
-		name: richRelationshipDefinition.name,
-		relationship: richRelationshipDefinition.relationship,
-		from: { ...from, otherNodeName: from.type },
-		to: { ...to, otherNodeName: to.type },
-		properties: richRelationshipDefinition.properties,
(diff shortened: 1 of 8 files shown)


packages/tc-schema-validator/validation/index.js
```diff
@@ -82,7 +82,7 @@ const signpost = error => {
 
 const fail = () => {
 	console.error('Treecreeper schema files invalid');
-	process.exit(2);
+	// process.exit(2);
```
GeoffThorpeFT, line 85: Temporary code whilst @nl-ria applies the validator upgrade - forces all yamls to be treated as valid.

packages/tc-schema-validator/validation/index.js
```diff
@@ -108,5 +108,5 @@ const fail = () => {
 
 process.on('unhandledRejection', error => {
 	console.error(error);
-	fail();
+	// fail();
```
GeoffThorpeFT, line 111: Temporary code whilst @nl-ria applies the validator upgrade - forces all yamls to be treated as valid.

packages/tc-schema-sdk/data-accessors/type.js
```diff
@@ -140,6 +145,7 @@ const createPropertiesWithRelationships = function ({
 					...updatedProps,
 					[propName]: {
 						...propDef,
+						...(propDef.cypher) && {otherNodeName: propDef.otherNodeName || propDef.type},   // only if its prop is a relationship
```
wheresrhys, line 148: I don't get why this should be here. This whole block only runs for non-relationships (cypher defined relationships are pseudo-non-relationships at this point as they don't need to deal with giving access to metadata on the relationship in the same way that ordinary relationship properties do)

packages/tc-schema-sdk/data-accessors/relationship-type.js
```diff
@@ -60,6 +62,17 @@ const formatSimpleRelationship = (
 	};
 };
 
+const formatRichRelationship = richRelationshipDefinition => {
+	const { from } = richRelationshipDefinition;
+	const { to } = richRelationshipDefinition;
```
wheresrhys, line 67:
```suggestion
	const { from, to } = richRelationshipDefinition;
```

packages/tc-schema-sdk/data-accessors/relationship-type.js
```diff
@@ -60,6 +62,17 @@ const formatSimpleRelationship = (
...
+	const enriched = {
+		...richRelationshipDefinition,
+		from: {...from, otherNodeName: from.type },
+		to: {...to, otherNodeName: to.type },
```
wheresrhys, line 71:
```suggestion
		from: {otherNodeName: from.type, ...from },
		to: {otherNodeName: to.type, ...to },
```

Assuming you want to default to whatever to/from provide (if defined) this is the patten to use. Properties get added from left to right, so this means the value you set here will get overridden by whatever already existed on to/from

packages/tc-schema-sdk/data-accessors/relationship-type.js
```diff
@@ -60,6 +62,17 @@ const formatSimpleRelationship = (
 	};
 };
 
+const formatRichRelationship = richRelationshipDefinition => {
```
wheresrhys, line 65: I've left some comments on the implementation, but I think this whole function could do with a rename and complete rewrite.

name: something like `setOtherNodeName`

implementation: I guess maybe part of the reason for implementing like this is because the objects are frozen so can't set a property on them? But even so, I think it'd help to check early  `if(from.otherNodeName && to.otherNodeName)` and just return richRelationshipDefinition without modification if it is. Then only carry out the thing you've implemented if necessary. This makes it far clearer that otherNodeName is sometimes defined already

wheresrhys:
Good progress. I think one function can be improved, and bit confused why one bit is doing what it is (feels like a hack workaround rather than a robust implementation at that point)

wheresrhys:
as discussed elsewhere, changes by neo4j to the entire neo4j-graphql direction mean this is no longer needed

PR #331, opened 2021-01-15
upgraded neo4j db and driver

wheresrhys:
## Why?

neo4j@3.x is EOL in september

## What?
Upgraded the obvious things that mention an explicit version number and/or a defined syntax change.

## Caveat
This PR does not cover any of the security changes (e.g. the introduction of the need for explicit certificates) as they will be tested by the calling clients - e.g. Biz Ops API


docker-compose.yaml
```diff
@@ -1,7 +1,7 @@
 version: '3'
```
GeoffThorpeFT, line 1: This should be `4` for consistency with the db version.

GeoffThorpeFT:
I have also tried to run this PR and as @wheresrhys suggests its broken. However, there is a clue which suggests we need to make a couple of further changes due to a change between v3 and v4 in the way database security works: https://neo4j.com/docs/driver-manual/current/client-applications/#driver-connection-uris

That's the good news :).   The bad news is that any db upgrade starting at v3 may need to go to v4.0 before it can then go to v4.1: https://neo4j.com/docs/migration-guide/4.0/

wheresrhys:
@GeoffThorpeFT v4.x will be backwards compatible with v4.0, so any advice for 4.0 should still apply

wheresrhys:
I can't approve this (as I created the PR 😝 ), but I think the next steps are to release a beta from this branch and then apply similar upgrades to Biz Ops API's codebase.

Then finally coordinating a release to biz ops test and to biz ops prod (Will be good to document how to upgrade the DB version via the new graphene DB console - the old docs in Biz Ops API will need updating)

Could you write up a few tickets for those please

GeoffThorpeFT:
With regard to the productionisation of this solution its important to note that the GrapheneDB solution to the certificate issue is to use a “new” protocol provided by neo4j.  They moved from bolt: to neo4j+s: as the URL prefix (scheme) to illustrate the use of a secure connection via a certificate.

GeoffThorpeFT:
https://github.com/neo4j/neo4j-javascript-driver/pull/534 as mentioned at the beginning of the driver release log for v4.0 https://github.com/neo4j/neo4j-javascript-driver/wiki/4.0-changelog

docker-compose.yaml
```diff
@@ -1,9 +1,9 @@
...
 services:
   neo4j:
-    image: neo4j:3.5.0
+    image: neo4j:4.1.5
     environment:
-      - NEO4J_dbms_logs_query_enabled=true
+      - NEO4J_dbms_logs_query_enabled=VERBOSE
```
GeoffThorpeFT, line 6: Slightly confused now with these settings as this v4.1 database can now be accessed by a v4 driver without the need for me to be explicit about the use of a certificate!!

packages/tc-api-db-manager/index.js
```diff
@@ -23,15 +23,13 @@ const initConstraints = async () => {
 		const retrieveConstraints = async () => {
 			const constraints = await executeQuery('CALL db.constraints');
 			return constraints.records.map(constraint =>
-				constraint.get('description'),
+				constraint.get('name'),
```
GeoffThorpeFT, line 26: This, and the similar line below, seem to be the solution to the original failure of the db to start. The reasons being that:
a) the description field is no longer available
b) the db.constraints and db.indexes procedures have been deprecated

docker-compose.yaml
wheresrhys: You're using `neo4j` protocol not `neo4j+ssc` https://github.com/Financial-Times/treecreeper/pull/331/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R20

Makefile
```diff
@@ -17,9 +17,9 @@ node_modules/@financial-times/rel-engage/index.mk:
 
 # End generated by rel-engage
 # ---------------------------
-LOCAL_BOLT_URL=bolt://localhost:7687
+LOCAL_BOLT_URL=neo4j://localhost:7687
 BIZ_OPS_BOLT_URL=${NEO4J_BOLT_URL}
-NEO4J_VERSION=3.5.0
+NEO4J_VERSION=4.1.5
```
wheresrhys, line 22: There is now 4.1.6

docker-compose.yaml
GeoffThorpeFT: Yes, I've now realised that at well, silly me.   However, this PR will now focus on the functionality delivered by the neo4j driver (e.g. CYPHER) not the security.   The security will be tested via a similar PR for Biz Ops API

Makefile
GeoffThorpeFT: Upgraded to 4.1.6 (all three places) - thanks

PR #339, opened 2021-02-18
start to write and use a more descriptive example schema

wheresrhys:
## Why?
The original idea with the test schema was to decouple tests from the arbitrariness (caused by the real world's peculiarities) of the Biz Ops model. So rather than e.g. using System-DELIVERED_BY->Team as the basis for many-to-one relationship tests, move to something more like AType-HAS_MANY_THINGS->AnotherType.

This itself had problems though
- The names chosen were often too abstract, and sometimes just plain bad
- Types were still reused for multiple things, so became bloated and confusing

## What?
Started introducing some better (I hope!) test types:

- KitchenSink - a type that has every kind of property on it (borrowing the name from the article that the publishing pipeline use to test)
- SimpleGraphBranch, SimpleGraphLeaf, FallenLeaf - implements a tree-like graph, with very few properties on it other than relationships, in order to test relationships. FallenLeaf should really be prefixed `SimpleGraph`, but as it is used to test properties relationships (the most complex bit of Biz Ops) I'm not sure yet if the choice of model will be an improvement when it comes to writing tests. Will do a find and replace when the decision is more settled.

Have also removed some abstractions from the tests - far fewer factory functions to create test data, and  ore explicit writing out objects in full. 

Have so far only updated tests for the api packages, so for now cannot delete the old schema.

Future PRs will tackle the e2e tests for the UI (which similarly suffer from overabstraction), at which stage the old schema can be removed


.circleci/config.yml
```diff
@@ -57,39 +58,16 @@ jobs:
...
+      # this ensures that the Cypress verified status is cached too
+      - run: make cypress-verify
 
       - *cache_node_modules
 
       - persist_to_workspace:
           root: *workspace_root
           paths:
             - .
-  cypress-build:
```
wheresrhys, line 67: Highlight #1 - removed a build step which was more or less just running npm install again + a cli tool to check the cypress binary got set up correctly (which I think is a piostinstall script in npm install anyway). Have brought that into the main install job and tweaked the statements of what gets put in the cache to makes sure the cypress binary is included

Makefile
```diff
@@ -132,6 +133,15 @@ cypress-primitives: build-statics
 
 ## Cypress stuff used in local dev
 
+run-app-stable:
```
wheresrhys, line 136: Highlight #2 When testing with cypress I found it good to have a server that doesn't keep restarting in response to code changes or the test runner never really gets off the ground.

TBH, the real solution is to tweak nodemon to ignore any changes to cypress test files or artefacts generated by cypress. Will do that before merging

cypress.json
```diff
@@ -1,5 +1,6 @@
 {
   "baseUrl": "http://localhost:8888",
   "integrationFolder": "./packages/tc-ui/src/",
-  "testFiles": "**/*.cyp.js"
+  "testFiles": "**/*.cyp.js",
+  "failOnStatusCode": false
```
wheresrhys, line 5: TODO - remove this - was just trying something out

example-schema/relationships/CuriousChild.yaml
```diff
@@ -1,38 +0,0 @@
-name: CuriousChild
```
wheresrhys, line 1: All these badly named schema files have gone

example-schema/type-hierarchy.yaml
```diff
@@ -2,11 +2,18 @@ test:
...
-    - MainType
-    - ParentType
-    - ChildType
     - RestrictedType
     - OddCodeType
     - MVRType
+    - MVRTypeChild
     - FieldsetType
     - ValidatedType
+    - PropertiesTest
```
wheresrhys, line 11: Highlight #4 there are now more types more directly tailored to testing specific things. Still not a perfect separation of concerns, but a big step in the right direction

packages/tc-api-rest-handlers/__tests__/get.spec.js
```diff
@@ -6,37 +6,36 @@ describe('rest GET', () => {
...
+		type: 'SimpleGraphBranch',
 		code: mainCode,
 	};
 
-	const { createNodes, createNode, connectNodes, meta } = setupMocks(
+	const { createNodes, createNode, connectNodes, stockMetadata } = setupMocks(
 		namespace,
 	);
 
-	const createMainNode = (props = {}) =>
```
wheresrhys, line 17: Highlight #5 Most of the clever helpers are gone. Left behind are some generic things like `createNode()` and `connectNodes()` but almost all wrapped version of these are gone in favour of actually just calling the generic with the actual full set of arguments in each test - slightly more verbose, but more comprehensible

packages/tc-api-rest-handlers/__tests__/absorb.spec.js
```diff
@@ -3,77 +3,62 @@ const { dbUnavailable } = require('../../../test-helpers/error-stubs');
...
-		jest.resetAllMocks();
-	});
+	const testPayload = {
+		type: 'SimpleGraphBranch',
+		code: mainCode,
+		codeToAbsorb: absorbedCode,
+	};
 
-	const absorb = absorbHandler({ documentStore });
-	const getInput = (override, metadata = {}) =>
```
wheresrhys, line 32: Highlight #6 Removes most (?all) these fancy test input factories in favour of defining some default payloads and using object spread to use these in tests (see 'green' line 23)

packages/tc-api-rest-handlers/__tests__/patch-create.spec.js
```diff
@@ -1,766 +1,57 @@
...
-const { dbUnavailable } = require('../../../test-helpers/error-stubs');
+const { setupMocks } = require('../../../test-helpers');
 
-const { patchHandler } = require('../patch');
+jest.mock('../post', () => ({
+	postHandler: jest.fn(),
+}));
+const { patchHandler: patchHandlerFactory } = require('../patch');
 
-describe('rest PATCH create', () => {
```
wheresrhys, line 6: Highlight #8 The tests in general don't stub anything out. However, when PATCHing a record that doesn't already exists it literally routes through ti the PST handler. To avoid the tedious duplication of tests (with the inevitable inconsistencies between what is tested in both cases) the test suite for patching non-existent records stubs out the post handler and just verifies it is called. This feels like a justifiable bit of stubbing.

packages/tc-api-rest-handlers/__tests__/patch-rich-relationships.spec.js
```diff
@@ -0,0 +1,405 @@
+const { setupMocks, neo4jTest } = require('../../../test-helpers');
```
wheresrhys, line 1: Highlight #9 All tests for patching rich relationships pulled into a separate file (previously they were sprinkled about a bit unnecessarily)_

packages/tc-markdown-parser/__tests__/nested-property-value.spec.js
```diff
@@ -8,14 +8,15 @@ schema.init({
...
-	type: 'MainType',
+	type: 'RelationshipTypeOne',
 });
 
 // The weird string concatenation in the markdown fixture is so that
 // dev tooling doesn't tidy up the 'extraneous' whitespace
 const addLineBreaks = str => str.replace(/\t+/g, '\t').split('\n').join('  \n');
 
-describe('nested property definition tests', () => {
+// we are soon to remove this functionality so not worried about updating tests to new schema
```
wheresrhys, line 18: Highlight #10 We'll soon be removing support for relationships in runbook.md so not bothering to refactor these tests to use the new types

packages/tc-schema-validator/validation/ad-hoc/relationship-consistency.js
```diff
@@ -90,12 +90,15 @@ const validateRelationshipConsistency = () => {
 			);
 
 			assert.equal(
-				fromPropDefs.length,
+				fromPropDefs.filter(
+					({ deprecationReason }) => !deprecationReason,
```
wheresrhys, line 94: Highlight #11 These changes allow for deprecating a relationship under one name and adding a new one that uses the same underlying data. The different test schema prompted this, but it is something that feels necessary: The tests should expect that active relationship properties come in pairs. but should be lenient on deprecated ones

packages/tc-ui/package.json
```diff
@@ -20,6 +20,9 @@
 		"react-dom": "^16.13.0",
 		"react-highlight-words": "^0.16.0"
 	},
+	"devDependencies": {
+		"neo4j-driver": "^4.1.2"
```
wheresrhys, line 24: because there is now a db fixtures creation helper in this package

packages/tc-ui/src/pages/delete/__tests__/e2e-delete-record.cyp.js
```diff
@@ -1,32 +1,22 @@
...
-	code,
-	deleteConfirmText,
-} = require('../../../test-helpers/mainTypeData.json');
-const {
-	createType,
-	visitEditPage,
-	visitMainTypePage,
-	pickChild,
-	save,
-} = require('../../../test-helpers/cypress');
```
wheresrhys, line 11: Highlight #12 These shared helpers wrapping collections of cypress commands are completely gone, replaced with a mixture of a handful of tailored helpers in each spec file to take care of the least DRY things, but mostly just done away with helpers, preferring to use cypress methods directly in tests

packages/tc-ui/src/pages/delete/__tests__/e2e-delete-record.cyp.js
```diff
@@ -1,32 +1,22 @@
...
-		cy.wrap(createType({ code, type: 'MainType' })).then(() => {
-			visitMainTypePage();
-		});
-	});
-
+	before(() => cy.wrap(dropFixtures(namespace)));
+	afterEach(() => cy.wrap(dropFixtures(namespace)));
 	it('shows a prompt message', () => {
-		cy.visit(`/MainType/${code}`, {
+		cy.wrap(
```
wheresrhys, line 11: Highlight #13 Previously a lot of fixtures were created _in_ cypress by navigating to a create page, saving the record etc. Now these are all created by using the `cy.wrap()` method to wrap direct db calls, which really speeds things up

packages/tc-ui/src/primitives/enum/__tests__/e2e-enum-primitive.cyp.js
```diff
@@ -0,0 +1,50 @@
...
+	});
+
+const createRecord = (props = {}) =>
+	cy.wrap(
+		executeQuery(`CREATE (:PropertiesTest $props)`, {
+			props: { code, ...props },
+		}),
+	);
+
+describe('End-to-end - Enum primitive', () => {
```
wheresrhys, line 18: Highlight #14 These specs for primitive components now follow a quite standard pattern
- view empty state (have uncovered that some components are buggy in this respect, so have raised issues and skipped tests for now)
- edit empty state
- add a value
- edit a value
- remove a value
- other things which will vary per component

The test suite fills me with more confidence now

PR #340, opened 2021-02-18
Add validation of DB credentials before initialising Neo4J driver

i-like-robots:
## Why?

Currently if credentials are missing or invalid the app fails to start with a cryptic error. 

Fixes https://github.com/Financial-Times/biz-ops-api/issues/504

## What?

Adds validation before initialising Neo4J driver in the DB manager package. After this change an explicit error is thrown indicating the problem.

No

## Screenshots / Images

Screenshot 2021-02-18 at 13 56 03


packages/tc-api-db-manager/db-connection.js
```diff
@@ -1,9 +1,23 @@
...
+// NOTE: This check ensures apps fail on startup instead of getting cryptic
+// error messages later.
+const validateEnvironment = () => {
+	assert.ok(process.env.NEO4J_BOLT_URL, 'Neo4J URL not set');
+	assert.match(process.env.NEO4J_BOLT_URL, /^neo4j\+ssc:/, 'Neo4J URL not valid');
+	assert.ok(process.env.NEO4J_BOLT_USER, 'Neo4J username not set');
+	assert.ok(process.env.NEO4J_BOLT_PASSWORD, 'Neo4J password not set');
+};
+
+if (process.env.NODE_ENV !== 'test') {
```
GeoffThorpeFT, line 17: You could probably run the assertions for the text env as well.

i-like-robots: I added this because the `bis-ops-api` tests were failing after making this change but I see the env vars for that app can be updated.

wheresrhys (approved):

PR #343, opened 2021-02-19
Add containsPII property to abstract type schema

i-like-robots:
## Why?

-   We want to ensure we omit fields potentially containing PII from logs or leaking into other systems
-   https://github.com/Financial-Times/biz-ops-schema/issues/324

## What?

-   Adds a new boolean property to abstract "node type" JSON schema
-   Updates docs to include the new property

I have linked this change into `biz-ops-schema` and tested amending an existing type to verify the change.


wheresrhys (approved):
LGTM :+1:

PR #346, opened 2021-02-22
Allow markdown in the descriptions and moreInformation for schema types.

GeoffThorpeFT:
## Why?

-   descriptions which included bullet points or formatting where being displayed on a single line without formatting

## What?

-   Added `markdownToHtml()` conversion before output of `description` or `moreInformation` in header aside.

no

### Screenshot
image

### DIFF
--- a/example-schema/types/MainType.yaml
+++ b/example-schema/types/MainType.yaml
@@ -1,7 +1,15 @@
 name: MainType
-description: Description of MainType.
+description: |
+  Description of MainType.
+  - bullet 1
+  - bullet 2
+  - bullet 3
+  Some more text
 moreInformation: |
   More information for MainType
+  - bullet
+  *bold text*
+  - another bullet
 properties:
   code:
     type: Code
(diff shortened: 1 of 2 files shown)


wheresrhys:
Is this the right solution? Should we rather edit the system description to move the markdowny bits into 'more information' (I think that was kinda the original intention anyway, which I had forgotten when implementing this change)

wheresrhys:
I stand corrected - markdown isn't supported for type descriptions _or_ type moreInformation anywhere as far as I can see.

There is a good reason to keep it to not supporting markdown though - if we know the content to be injected somewhere will only ever be a paragraph that means we can make a lot of assumptions when outputting and styling it. Once we allow markdown we are basically saying 'this could contain _any_ html', which will lead to headaches.

My proposal: 
- description should not allow markdown
- moreInformation should allow markdown
- within the Biz Ops ecosystem we should never output description without an expander to show moreInformation

We need to update everywhere that description is output to conform to these rules.

_but_ in the short term I don't think it's a serious enough problem to warrant that level of overhaul. No-one outside the team has even noticed yet.

wheresrhys:
Have raised an issue to capture the problem https://github.com/Financial-Times/treecreeper/issues/361

PR #347, opened 2021-02-24
Bug fix for views without properties in props

katsuraku:
## Why?

-   Bug issue [here](https://github.com/Financial-Times/biz-ops-admin/issues/554)

## What?

-   Check whether properties are there first - this tweak made inline in biz-ops-admin, and run locally, fixes the bug reported on the list view


wheresrhys:
I'd rather fix this in Biz Ops admin because what this change done here effectively says is 'in all cases where no properties exist, don't error - just skip the task of displaying them silently' - that might not always be the right thing to do.

PR #353, opened 2021-03-08
delete properties rather than storing empty strings

wheresrhys:
## Why?

Biz Ops could not unset any strings stored in Document properties

Fixes https://github.com/Financial-Times/biz-ops-admin/issues/548
## What?

- There was some code explicitly preventing passing empty strings to the s3 document store. While this has good intentions - don't want to create documents when we have no data to write - it's premature and means that those empty properties don't get through to the docstore in situations where the user wants to delete something. This code has been removed and the test updated to expect that a null gets passed through
- Diffing logic in the docstore has been updated to avoid deciding a write is needed when comparing `null` (from the API) with `undefined` (from S3), 
- Before writing to S3 the docstore deletes any falsy values from the body, which effectively deletes them from s3

TODO

- [x] test to make sure does not write when posting an object just containing an empty string

### DIFF
--- a/packages/tc-api-rest-handlers/lib/separate-documents-from-body.js
+++ b/packages/tc-api-rest-handlers/lib/separate-documents-from-body.js
@@ -1,4 +1,3 @@
-const _isEmpty = require('lodash.isempty');
 const { getType } = require('@financial-times/tc-schema-sdk');
 
 const separateDocsFromBody = (nodeType, body = {}) => {
@@ -8,14 +7,9 @@ const separateDocsFromBody = (nodeType, body = {}) => {
 
 	Object.entries(body).forEach(([key, value]) => {
 		if (key.charAt(0) !== '!' && properties[key].type === 'Document') {
-			// We should check the value is empty only when type is Document
-			// because other type will be passed as boolean, number, etc...
-			// then _isEmpty() will returns true
-			if (!_isEmpty(value)) {
-				bodyDocuments[key] = body[key];
-			}
+			bodyDocuments[key] = value;
 		} else {
-			bodyNoDocs[key] = body[key];
+			bodyNoDocs[key] = value;
 		}
 	});
 
--- a/packages/tc-api-rest-handlers/lib/separate-documents-from-body.js
+++ b/packages/tc-api-rest-handlers/lib/separate-documents-from-body.js
@@ -6,7 +6,7 @@ const separateDocsFromBody = (nodeType, body = {}) => {
 	const bodyNoDocs = {};
 
 	Object.entries(body).forEach(([key, value]) => {
-		// checking for `!` avoids errors being thrown when iterating over 
+		// checking for `!` avoids errors being thrown when iterating over
 		// 'delete relationships' properties
 		if (key.charAt(0) !== '!' && properties[key].type === 'Document') {
 			bodyDocuments[key] = value;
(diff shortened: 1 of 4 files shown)


packages/tc-api-rest-handlers/lib/separate-documents-from-body.js
```diff
@@ -8,14 +7,9 @@ const separateDocsFromBody = (nodeType, body = {}) => {
 
 	Object.entries(body).forEach(([key, value]) => {
 		if (key.charAt(0) !== '!' && properties[key].type === 'Document') {
-			// We should check the value is empty only when type is Document
-			// because other type will be passed as boolean, number, etc...
-			// then _isEmpty() will returns true
-			if (!_isEmpty(value)) {
-				bodyDocuments[key] = body[key];
-			}
+			bodyDocuments[key] = value;
```
GeoffThorpeFT, line 10: Should there be a check for '!' and type = Document?  Currently it just falls through to the else.

packages/tc-api-s3-document-store/patch.js
```diff
@@ -11,8 +11,13 @@ const s3Patch = async ({ s3Instance, bucketName, type, code, body }) => {
 		code,
 	});
 	const changedProperties = Object.keys(body).filter(
-		key => body[key] !== existingBody[key],
+		// check that at least one has a value to avoid e.g. null !== undefined causing
+		// an unnecessary write (the api will send nulls sometimes, but the json object will
+		// contain undefineds)
+		key =>
+			(body[key] || existingBody[key]) && body[key] !== existingBody[key],
```
GeoffThorpeFT, line 18: Does the && belong to the !== existing key or the OR?  Some extra brackets would help my understanding.

packages/tc-api-s3-document-store/patch.js
```diff
@@ -28,10 +33,15 @@ const s3Patch = async ({ s3Instance, bucketName, type, code, body }) => {
 	}
 
 	const newBodyDocument = Object.assign(existingBody, body);
+
+	const newBodyWithoutEmpty = Object.fromEntries(
+		Object.entries(newBodyDocument).filter(([, value]) => !!value),
+	);
+
```
GeoffThorpeFT, line 40: Great. This was the piece of code I was looking for when I saw the bug.  Its much clearer that the new body should exclude blank/null/undefined fields.

packages/tc-api-s3-document-store/patch.js
wheresrhys: I agree, but our linting rules will convert it back to this though (somethig like 'no extraneous brackets')

packages/tc-api-rest-handlers/lib/separate-documents-from-body.js
wheresrhys: Starting with `!` is only valid for relationships, so having both together isn't possible. Happy to add a comment though

packages/tc-api-rest-handlers/lib/separate-documents-from-body.js
```diff
@@ -8,14 +7,9 @@ const separateDocsFromBody = (nodeType, body = {}) => {
 
 	Object.entries(body).forEach(([key, value]) => {
 		if (key.charAt(0) !== '!' && properties[key].type === 'Document') {
```
wheresrhys, line 9:
```suggestion
		// checking for `!` avoids errors being thrown when iterating over 
		// 'delete relationships' properties
		if (key.charAt(0) !== '!' && properties[key].type === 'Document') {
```

PR #354, opened 2021-03-16
Footers

hdsft:


GeoffThorpeFT (changes requested):
This doesnt look like a change that is necessary.  The only difference is the spacing.
It looks like the hardcoding had already been removed by someone else.

GeoffThorpeFT:
I'm going to close this PR as its just a change to spacing

PR #355, opened 2021-03-31
POC of releasing beta features in one schema

wheresrhys:
## Why?

2 motivations:

- be able to publish a schema change to test without a risk of it getting to prod, while having it persist reliably in test
- be able to add test definitions to the schema that can be used for writing comprehensive tests of various bits of the ecosystem (in particular UI components)

## What?
Various aspects of the schema (types, enums, properties, ...) may have an `isTest: true` property. 

The schema SDK can be initialised with an `includeTest` flag. If false these test types are stripped out.

This option can be set by `tc-schema-publisher` as a command line argument

Thus a single schema can be authored to include test elements and published with the flag set to `true` to the test environment and with the flag set to `false` for the prod environment.

The workflow for trying out schema changes will be to author with the `isTest` flag set. Once we're satisfied that it works in test we can remove the flags and publish to prod.

The property filtering happens in the `RawDataWrapper` bit of tc-schema-sdk, so at the lowest level, before the data is transformed by the rest of the SDK for use by the developer. This feels a bit icky - it means `RawData` doesn't exactly mean `RawData` - but feels like the safest place to do it to avoid accidental leaking of test types into prod. It also means that all the filters are located in a single file, rather than distributed around all the [sometimes complex] data-accessor functions

## TODO
- [x] tc-schema-sdk needs tests
- [ ] the filters' behaviour needs tests too


GeoffThorpeFT:
It is great to see the use of `isTest` throughout the schema files.

Does this mean that a new type of data would need `isTest` against every property or could `isTest` be defined once at the top of the yaml?

Does the hierarchy file also need `isTest` to allow new test types to be added?

packages/tc-schema-sdk/lib/raw-data-wrapper.js
```diff
@@ -39,7 +87,8 @@ If npm linking the schema locally, set \`updateMode: 'dev'\`
 
 	getEnums() {
 		this.checkDataExists();
-		return this.rawData.schema.enums;
+		// TODO - allow test enum values
```
GeoffThorpeFT, line 90: Do test enums really impact the prod env? Feels like we could just let them be deployed without triggering any issues.

wheresrhys: say we add a new value to the lifecycleStage enum 'Planning' - it'd mean that would show up in both test and prod. I don't think it's a major issue, hence TODO

wheresrhys:
> Does this mean that a new type of data would need isTest against every property or could isTest be defined once at the top of the yaml?

If it's a new type then once at the top of the file will suffice. For existing types add per property to hide individual bits.

> Does the hierarchy file also need isTest to allow new test types to be added?

Nope, all types (test and prod) should be referenced, and the SDK takes care of cleaning out any that shouldn't be there based on `isTest` being present in the individual type files

TODO = DOCUMENTaTION!!!

wheresrhys:
closing in favour of the slightly neater https://github.com/Financial-Times/treecreeper/pull/357

PR #365, opened 2021-06-22
Implement option to make writes more efficient

wheresrhys:
## Why?

Fixes this https://github.com/Financial-Times/biz-ops-api/issues/534

## What?

When fetching data from the database (both the preflight request for diffing and the final response to user) only retrieves relationships relevant to the current request


kiyaGu (approved):
I think I kind of need to refresh my knowledge of treecreeper but from a high level overview it LGTM 👍🏾

PR #366, opened 2021-06-25
Disable the Code field during edit

GeoffThorpeFT:
## Why?

-   Users can edit the `Code` field but if they do they receive an unhelpful message.

## What?

-   Disable the `Code` field during edit
-  Display a helpful message about the use of `absorb` to achieve a code rename

No

image

### DIFF
--- a/packages/tc-ui/src/primitives/text/server.jsx
+++ b/packages/tc-ui/src/primitives/text/server.jsx
@@ -9,6 +9,7 @@ const localOnChange = (event, onChange) => {
 };
 
 const EditText = ({
+	isEdit,
 	propertyName,
 	value,
 	required,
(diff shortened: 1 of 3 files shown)


packages/tc-ui/src/primitives/text/server.jsx
```diff
@@ -9,35 +9,54 @@ const localOnChange = (event, onChange) => {
...
+	type,
 	onChange,
 }) => {
 	const name = !isNested
 		? `${propertyName}${lockedBy || disabled ? '-disabled' : ''}`
 		: '';
 
+	// Special case for locking codes during edit
+	let additionalDescriptiveText;
+	if (type === 'Code' && isEdit) {
```
wheresrhys, line 29: Can you go with `propertyName === 'code'` as the check. `type: 'Code'` used to power a few things, but I'm not sure it does any more and I'm not sure it's worth keeping around really.

wheresrhys (commented):
This is totally the right fix, but in the wrong code base - Biz Ops Admin now contains its own set of components (in a future edition of treecreeper I will remove tc-ui).

I should spend some time moving issues from treecreeper to other repos as appropriate

